PR #7, opened 2023-08-24
Fix compatibility with new BitStream 

medley56:
Shows up when filling our BitStream ring buffer for reading large binary files in chunks. Previously, += added new data to the buffer but left the cursor (pos) in place. New behavior moves the cursor to the end of the BitStream. This fix forces that cursor to stay put when filling the ring buffer.

### DIFF
--- a/space_packet_parser/parser.py
+++ b/space_packet_parser/parser.py
@@ -445,24 +445,27 @@ class PacketParser:
                 Number of bits added to the buffer. Note that the buffer may still have nonzero length from previous
                 data even when this returns zero.
             """
+            curser_pos = buffer.pos  # Keep track of the original buffer cursor location
             if isinstance(source, io.BufferedIOBase):
                 new_bytes = source.read(read_size_bytes)
                 buffer += new_bytes
-                return len(new_bytes)*8
-
-            if isinstance(source, socket.socket):
+                n_new_bits = len(new_bytes)*8
+            elif isinstance(source, socket.socket):
                 new_bytes = source.recv(read_size_bytes)
                 buffer += new_bytes  # Append BitStream with newly read bytes
-                return len(new_bytes)*8
-
-            if isinstance(source, bitstring.ConstBitStream):
+                n_new_bits = len(new_bytes)*8
+            elif isinstance(source, bitstring.ConstBitStream):
+                # This either reads read_size_bytes bytes or it just reads to the end of the data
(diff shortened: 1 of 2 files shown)


space_packet_parser/parser.py
```diff
@@ -445,24 +445,27 @@ def fill_read_buffer(source: bitstring.ConstBitStream or BinaryIO or socket.sock
                 Number of bits added to the buffer. Note that the buffer may still have nonzero length from previous
                 data even when this returns zero.
             """
+            curser_pos = buffer.pos  # Keep track of the original buffer cursor location
```
greglucas, line 448:
```suggestion
            cursor_pos = buffer.pos  # Keep track of the original buffer cursor location
```

space_packet_parser/parser.py
```diff
@@ -445,24 +445,27 @@ def fill_read_buffer(source: bitstring.ConstBitStream or BinaryIO or socket.sock
...
+                source.pos += len(new_bits)  # Set the source.pos to exactly where we read to
                 buffer += new_bits
-                return len(new_bits)
+                n_new_bits = len(new_bits)
+            else:
+                raise ValueError(f"Unrecognized data source: {source}")
 
-            raise ValueError(f"Unrecognized data source: {source}")
+            # Reset buffer.pos to the original position before we extended it
+            buffer.pos = curser_pos
```
greglucas, line 467: Is the += always adding n_new_bits to the position? If so, could you just remove that at the end here instead of needing to keep track of the original value?
```suggestion
            buffer.pos -= n_new_bits
```

greglucas (commented):
It seems like you'd want to move the pointer with the data when you're adding to the buffer? So, I guess that change makes intuitive sense to me, but maybe I'm missing something...

I'm wondering if you could move this logic up a level to keep track of the last cursor position in the reading loops instead of relying on this routine to fill data without moving the cursor?

space_packet_parser/parser.py
medley56: It seems more reliable to book keep the original cursor location than to do math to try to reconstruct the original location later on. Both would work though I'm sure.

space_packet_parser/parser.py
medley56: Yeah, I understand why the Bitstring author made the change. This just makes my code behave the same way the old Bitstring worked because we fundamentally need the ringbuffer cursor not to move just because we added some data to it.

PR #8, opened 2023-08-24
Add docs for common issues

medley56:


docs/source/users.md
```diff
@@ -70,6 +70,17 @@ These exception objects are not raised so that the generator may keep parsing. I
...
+## Common Issues and Solutions
+
+### Parser Generator Completes without Yielding a Packet
+This can occur if your data file contains only packets that do not match any packet definitions in your XTCE document 
+and `yield_unrecognized_packet_errors=False` (the default). This could mean that your data file actually contains only 
+APIDs that are not covered in your packet definition but usually it means you have bad restriction criteria.
+
+### Only Packet Headers are Parsed
+If you observe that only packet headers are being parsed but no exceptions are being raised, it likely means that 
+you have forgotten to set `abstract="true"` on your non-concrete sequence container elements.
```
greglucas, line 82: Give a concrete example of what you mean by the elements to update?
`<... abstract="true">`

docs/source/users.md
```diff
@@ -70,6 +70,17 @@ These exception objects are not raised so that the generator may keep parsing. I
 are yielded from the generator with a `partial_data` attribute for user examination. This partial data allows you to 
 see how far it got through a container inheritance structure before failing to determine the rest of the structure. 
 
+## Common Issues and Solutions
+
+### Parser Generator Completes without Yielding a Packet
+This can occur if your data file contains only packets that do not match any packet definitions in your XTCE document 
+and `yield_unrecognized_packet_errors=False` (the default). This could mean that your data file actually contains only 
+APIDs that are not covered in your packet definition but usually it means you have bad restriction criteria.
```
greglucas, line 78: Is there a concrete example for what "bad restriction criteria" means as well? I think spelling these things out explicitly with one example will allow people to generalize easier.

greglucas (commented):
FYI, this commit added new binary data that is unrelated. I suggest removing that file from the commit unless it is needed/used somewhere.

greglucas:
I really like those examples, I think that makes it much clearer what is required for end users!

PR #9, opened 2023-08-24
Add error handling to the textio error

medley56:
Documentation of example usage has been updated with proper "rb" mode.
Exception handling has been added in the event that a file is opened in non-binary mode.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches: [ "master", "dev" ]
   pull_request:
-    branches: [ "master" ]
+    branches: [ "master", "dev" ]
 jobs:
   python-version-matrix:
     strategy:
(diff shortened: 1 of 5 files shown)


greglucas (approved):
I don't get why half of the tests are running for hours here... even on your other documentation PR.

medley56:
@greglucas Until the Bitstring bugfix is merged, none of the other PRs will pass their checks because those checks are pulling in the most recent Bitstring library.

greglucas:
> @greglucas Until the Bitstring bugfix is merged, none of the other PRs will pass their checks because those checks are pulling in the most recent Bitstring library.

Ahh, I did not catch that the other tests were pinned back. It is unfortunate that the bug "hangs" rather than raising an error and failing the process. We might want to add a shorter timeout to the actions runners so we don't waste CPU on the CI? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

PR #15, opened 2023-11-03
Remove upper pin restriction on the package

greglucas:
This allows installation with Python 3.12+ and newer versions of bitstring as well.

homepage also must be a uri and not an empty string, so I added that in.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,8 +8,8 @@ jobs:
   python-version-matrix:
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
-        bitstring-version: ["", "3.0.0"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        bitstring-version: ["", "3.0.0", "4.1.2"]
     uses: ./.github/workflows/test-python-version.yml
     with:
       python-version: ${{ matrix.python-version }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ maintainers = [
     "Gavin Medley <gavin.medley@lasp.colorado.edu>"
 ]
 repository = "https://github.com/medley56/space_packet_parser"
-homepage = ""
+homepage = "https://space-packet-parser.readthedocs.io"
 keywords = [
     "ccsds",
     "xtce",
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -34,8 +34,8 @@ keywords = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.12"
-bitstring = ">=3.0.0,<4.2"
```
medley56, line 38: I had bitstring pinned because last time I allowed bitstring to upgrade through minor versions, it introduced a bug because they changed their API on a minor release. If you add a matrix test that includes the latest version of bitstring, we can update the high pin but I don't trust bitstring not to introduce breaking changes anymore.

.github/workflows/tests.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   python-version-matrix:
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         bitstring-version: ["", "3.0.0"]
```
medley56, line 12: i.e. add a 4.1.2 version here

pyproject.toml
```diff
@@ -13,7 +13,7 @@ maintainers = [
     "Gavin Medley <gavin.medley@lasp.colorado.edu>"
 ]
 repository = "https://github.com/medley56/space_packet_parser"
-homepage = ""
+homepage = "https://space-packet-parser.readthedocs.io"
```
medley56, line 16: Thanks!

medley56 (changes requested):
Please add a matrix test for the latest version of bitstring.

pyproject.toml
greglucas: I hear you on being frustrated with an upstream package doing that, but I'm not sure narrowing the version window on this package is a great solution. This means you'll need to keep iterating and re-releasing every time new versions come out of either of these. I would rather not have to move the upper pins and re-release every time a new version comes out that I want to support, but that of course requires one to deal with any fallout if something upstream does break the package. So pick your poison I guess.

I think the upper pin is to the most recent version now, so what would you prefer the upper pin to be? `<5` and assume the major version wouldn't work? Or you're saying we can remove the upper pin and just add a `4.1.2` test below? (I think the `""` gets the latest in the test suite?)

greglucas: @medley56 What is the change you'd like to see for the bitstring pin, or is it OK to leave it unpinned and just add the "4.1.2" version to the test matrix which I just added and pushed up.

medley56: We can leave it unpinned and see if the breakage happens again. There isn't a perfect solution to other maintainers introducing breaking changes in minor versions.

medley56 (approved):
Let's get this merged.

PR #20, opened 2024-01-25
Add more `*ParameterType` classes

cgobat:
Straightforward PR that adds the following classes:
- `BooleanParameterType`
- `AbsoluteTimeParameterType`
- `RelativeTimeParameterType`

This is a rehash of #19 that includes the latter two time classes as well. As I noted there, none of these should need any additional encoding definitions, as they would all typically use an `IntegerDataEncoding` or `FloatDataEncoding`.

### DIFF
--- a/space_packet_parser/xtcedef.py
+++ b/space_packet_parser/xtcedef.py
@@ -1742,6 +1742,11 @@ class IntegerParameterType(ParameterType):
     pass
 
 
+class BooleanParameterType(ParameterType):
+    """<xtce:BooleanParameterType>"""
+    pass
+
+
 class FloatParameterType(ParameterType):
     """<xtce:FloatParameterType>"""
     pass
@@ -1948,6 +1953,7 @@ class XtcePacketDefinition:
     _tag_to_type_template = {
         '{{{xtce}}}StringParameterType': StringParameterType,
         '{{{xtce}}}IntegerParameterType': IntegerParameterType,
+        '{{{xtce}}}BooleanParameterType': BooleanParameterType,
         '{{{xtce}}}FloatParameterType': FloatParameterType,
         '{{{xtce}}}EnumeratedParameterType': EnumeratedParameterType,
         '{{{xtce}}}BinaryParameterType': BinaryParameterType,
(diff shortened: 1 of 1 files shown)


medley56:
I think there is more work to be done to fully capture the encoding of time types.
See Section 9.2.1.3 of https://public.ccsds.org/Pubs/660x1g2.pdf for some examples of time parameter types with encodings. Note that the encodings there are wrapped in an `xtce:Encoding` element. This differs from the way that other parameter type encodings are represented. As a result, I believe attempting to parse an AbsoluteTimeParameterType element will fail because it won't find the encoding. 

My second request is to add tests that demonstrate correct parsing of each of the parameter types you are adding from XML excerpts. See `test_xtcedef.py` for examples of such tests for other parameter types. Please continue the pattern of naming tests and the pattern of parameterizing them with various different valid XML excerpts to fully test the functionality.

cgobat:
I think commit 90dc80c should already (mostly?) address your first point. I'm using `//` XPath syntax to look for encoding elements at all levels beneath each parameter type element, rather than just direct children. That means it will look for and find a `*DataEncoding` element regardless of whether it is immediately beneath the `*ParameterType` or is nested another level down within a dedicated `Encoding` element.

I have used this modified version to parse XTCE/packets using `AbsoluteTimeParameterType` fields, and it's working fine for me. I can work on adding some related tests, but if you tried it and actually encountered a problem, let me know.

medley56:
@cgobat I'm going to take a stab at writing tests for this because I want to support these types, especially if SWRI is using them! But I'm not willing to add code to the library that isn't covered by unit tests.

cgobat:
Yep, makes sense. Sorry I haven't had time to get back to this in the last few weeks.

medley56:
@cgobat I believe I've incorporated these changes. Can you try using version 4.2.0rc2 and let me know if you have any issues? It's up on PyPI.

cgobat:
I had a chance to test out 4.2.0rc2, and everything looks good with respect to the new time parameter types. I did re-encounter some endianness issues, but I think that's to be expected and is what reminded/partially inspired me to open #26.

cgobat:
I believe the re-implementation of these changes in #24 makes this PR essentially a duplicate now. Should this be closed?

medley56:
Closing this as the features requested have been implemented in PR #24. Thanks @cgobat for getting this work going. Your work was very helpful!

PR #22, opened 2024-02-18
PERF: Reduce number of buffer resize operations

greglucas:
Previously the buffer was trimmed after every packet parsed. Now we only trim every 20 MB read.

* Would we want to make the 20 MB user controllable via keyword argument? (I just arbitrarily chose that value)
* Question for the future: Would we want to default to reading the entire files / draining the sockets now? It would probably be good to make the file read size larger than 4096 so there aren't as many appends happening, but we also might not want to read an entire file into memory, so perhaps there is a sweet spot for the default.

Here is an example that is helped by this. I'm not sure there is a great test we can add, but open to thoughts/ideas on that. I think the problem in this case is that we read everything into a large BitStream and then keep trimming the front of the array which is expensive for large bitstreams (making a new large bitstream every time it slices a small chunk off I believe).

```python
from pathlib import Path
import time

from space_packet_parser import parser, xtcedef

# Create a large file to test the performance of the parser
# quick bash script to append idex a few times
# for i in {1..100};do cat tests/test_data/idex/sciData_2023_052_14_45_05 >> idex-many-files.bin; done

p = Path("tests/test_data/idex")

xtce_document = p / "idex_combined_science_definition.xml"
packet_file = Path.cwd() / "idex-many-files.bin"

packet_definition = xtcedef.XtcePacketDefinition(xtce_document)

idex_parser = parser.PacketParser(packet_definition)

start = time.time()
i = 0
with open(packet_file, "rb") as binary_data:
    idex_packets = idex_parser.generator(
        binary_data,
        # This will read the full file every read. Compare main branch
        # to this branch to see the performance difference
        # This branch: Parsed 7800 packets in 5.77 seconds: 1352.11 packets/s
        # Main branch: Parsed 7800 packets in 36.71 seconds: 212.50 packets/s
        # Main branch (default buffer read size): Parsed 7800 packets in 6.07 seconds: 1285.83 packets/s
        buffer_read_size_bytes=-1,
    )
    for packet in idex_packets:
        i += 1
        pass
total_time = time.time() - start
print(f"Parsed {i} packets in {total_time:.2f} seconds: {i/total_time:.2f} packets/s")
```

### DIFF
--- a/space_packet_parser/parser.py
+++ b/space_packet_parser/parser.py
@@ -501,8 +501,15 @@ class PacketParser:
                 self.print_progress(current_bits=n_bits_parsed, total_bits=total_length_bits,
                                     start_time_ns=start_time, current_packets=n_packets_parsed)
 
+            start_pos = read_buffer.pos
+            if start_pos > 160_000_000:
+                # Only trim the buffer after 20 MB read to prevent modifying
+                # the bitstream and trimming after every packet
+                read_buffer = read_buffer[start_pos:]
+                start_pos = 0
+
             # Fill buffer enough to parse a header
-            while len(read_buffer) < skip_header_bits + CCSDS_HEADER_LENGTH_BITS:
+            while len(read_buffer) - start_pos < skip_header_bits + CCSDS_HEADER_LENGTH_BITS:
                 result = fill_read_buffer(binary_data, read_buffer,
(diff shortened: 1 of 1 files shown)


medley56 (approved):
Yep. I think this makes sense. All tests pass. The example still works. LGTM.

PR #25, opened 2024-03-15
PERF: Use bytes objects for the buffer rather than bitstring

### LINKED ISSUE #33: Replace main parsing buffer with a bytes object
When reading a file, the main buffer is currently a BitStream object. This is inefficient compared to a bytes object buffer. Replace the top level buffer with a bytes object (we can rely on all packets being an integer number of bytes overall).

greglucas:
# Title of PR

This uses bytes objects directly in the buffers. Keeping track of the current position manually. Currently, this provides a modest speedup for full packet parsing because we still create a new ConstBitStream for each packet parsing. But for `ccsds_header_only=True` we get a 10x speedup.

I'm opening this as a draft for now to aid discussion.

The other thing that helped with the speed here is defaulting to a full read of a file rather than `file.read(4096)` as it was before and creating lots of small appends to the bytes buffer. Happy to remove this if you'd prefer people to input -1 manually to get that, but figured it is likely what people would want (may be wrong there).

closes #33

## Checklist

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -17,6 +17,7 @@ Release notes for the `space_packet_parser` library
 - Add support for BooleanParameterType
 - Drop support for bitstring <4.0.1
 - Support BooleanExpression in a ContextCalibrator
+- Default read size is changed to a full file read on file-like objects
 
 ### v4.1.1 (released)
 - Allow Python 3.12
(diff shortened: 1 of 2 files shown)


medley56:
@greglucas Do you want to polish this PR? It's currently a draft.

greglucas:
I had it in draft so we could discuss, but after yesterday's discussion I think this should be good for review now. The major external facing change is noted in the changelog and is that we read in the entire file as a full byte read by default now.

medley56 (approved):
Nice optimization!

PR #26, opened 2024-03-18
Add straightforward endianness implementation; accommodate all allowable string encodings

cgobat:
This PR ended up turning into a kind of rollup of several of the changes I've made in my fork to make the module usable. I am opening this as a draft for discussion purposes and as a sort of already-implemented feature(s) request. The major changes here are:
- ~Addition of a `BooleanParameterType` class~
- Implementation of byte ordering/endianness checking
  - While in an ideal world, spacecraft subsystems, etc. would all stick to big endian data types, in practice that is not the case.  Hardcoding big endian types for all parameters is not a safe assumption, and the ability to handle both MSB- and LSB-first fields is a pretty fundamental/core functionality
  - I noticed several notes in the code along the lines of `TODO: implement ByteOrderList to inform endianness.` I'm not sure what "ByteOrderList" is a reference to (a name idea for a new/future class?), but the way I've done it here is pretty clean (IMO) and seems to be fully transparent and backwards compatible in my limited testing. Happy to discuss if you had something else in mind for this though.
- Addition of the complete set of XTCE-allowable character encoding values for `StringDataEncoding`s, plus integration with the aforementioned endianness parsing for the cases where byte order is described by the string encoding

## Checklist
  - Note: I still think it is worth it to add an `lxml` dependency, but I will leave that for another PR

### DIFF
--- a/space_packet_parser/xtcedef.py
+++ b/space_packet_parser/xtcedef.py
@@ -1180,30 +1180,33 @@ class DataEncoding(AttrComparable, metaclass=ABCMeta):
 class StringDataEncoding(DataEncoding):
     """<xtce:StringDataEncoding>"""
 
-    def __init__(self, encoding: str = 'utf-8',
-                 termination_character: Optional[str] = None,
-                 fixed_length: Optional[int] = None,
-                 leading_length_size: Optional[int] = None,
-                 dynamic_length_reference: Optional[str] = None,
+    _supported_encodings = ('US-ASCII', 'ISO-8859-1', 'Windows-1252', 'UTF-8', 'UTF-16',
+                            'UTF-16LE', 'UTF-16BE', 'UTF-32', 'UTF-32LE', 'UTF-32BE')
+
+    def __init__(self, encoding: str = 'UTF-8',
+                 byte_order: str = None,
+                 termination_character: str = None,
+                 fixed_length: int = None,
+                 leading_length_size: int = None,
+                 dynamic_length_reference: str = None,
                  use_calibrated_value: bool = True,
-                 discrete_lookup_length: Optional[List[DiscreteLookup]] = None,
-                 length_linear_adjuster: Optional[callable] = None):
-        """Constructor
+                 discrete_lookup_length: list = None,
+                 length_linear_adjuster: callable = None):
+        f"""Constructor
         Only one of termination_character, fixed_length, or leading_length_size should be set. Setting more than one
         is nonsensical.
-
-        TODO: implement ByteOrderList to inform endianness.
-         This can also relax the requirements on the encoding spec since utf-16-le is redundant if endianness
-         comes from the ByteOrderList
 
         Parameters
         ----------
         encoding : str
-            One of 'utf-8', 'utf-16-le', or 'utf-16-be'. Describes how to read the characters in the string.
-        termination_character : Optional[str]
+            One of {self._supported_encodings}. Describes how to read the characters in the string.
+        byte_order : str
+            Description of the byte order, used for multi-byte character encodings where the endianness cannot be
+            determined from the encoding specifier. Can be None if encoding is single-byte or UTF-*BE/UTF-*LE.
+        termination_character : str
             A single hexadecimal character, represented as a string. Must be encoded in the same encoding as the string
             itself. For example, for a utf-8 encoded string, the hex string must be two hex characters (one byte).
-            For a utf-16-* encoded string, the hex representation of the termination character must be four characters
+            For a UTF-16* encoded string, the hex representation of the termination character must be four characters
             (two bytes).
         fixed_length : Optional[int]
             Fixed length of the string, in bits.
@@ -1220,10 +1223,8 @@ class StringDataEncoding(DataEncoding):
             Function that linearly adjusts a size. e.g. if the size reference parameter gives a length in bytes, the
             linear adjuster should multiply by 8 to give the size in bits.
         """
-        if encoding not in ['utf-8', 'utf-16-le', 'utf-16-be']:
-            raise ValueError(
-                f"Got encoding={encoding}. Encoding must be one of utf-8, utf-16-le, or utf-16-be (note that"
-                f"endianness must be specified for utf-16 encoding.")
+        if encoding not in self._supported_encodings:
+            raise ValueError(f"Got encoding={encoding}. Encoding must be one of {self._supported_encodings}.")
         self.encoding = encoding
         # Check that the termination character is a single character in the specified encoding
         # e.g. b'\x58' in utf-8 is "X"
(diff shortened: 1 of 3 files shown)


medley56:
@cgobat This is great work and I want to get these changes in. I think I will do the same thing as your last PR and take these changes and incorporate them into a different changeset that includes tests and omits a few pieces (e.g. I'm still torn on incorporating the lxml dependency and a lot of tests will need to be reworked for that I think). If you're concerned about claiming "credit" for the changes, I can try to teach myself how to continue work on your branch. Is it as simple as me making commits to your fork?

The BooleanParameterType was part of PR #24, along with the time parameter types so I think we'll pull those changes out of this PR.

On the subject of the ByteOrderList element, the XTCE green book element description is a little unclear on it. I think in XTCE 1.1 there was a ByteOrderList element that literally spelled out the byte significant for every byte in a data encoding. However, it looks like in 1.2 this may have been superseded by the mostSignificantByteFirst and leastSignificantByteFirst attributes that you are using. The explicit element-form of ByteOrderList appears to be still allowed in XTCE 1.2 but seems like an extremely niche use-case now that we have the attribute forms instead. In short, I think your implementation is sufficient.

See https://issues.omg.org/issues/XTCE12-400 for a discussion of the byte ordering changes between XTCE 1.1 and 1.2.

cgobat:
> If you're concerned about claiming "credit" for the changes, I can try to teach myself how to continue work on your branch. Is it as simple as me making commits to your fork?

It's not a huge deal to me. It might be nice, but only if it's not too inconvenient for you and your workflow. That said, GitHub does make it pretty easy to continue working on this branch&mdash;just `git fetch origin pull/26/head:pr26` (assuming your remote is called `origin`; you can change `pr26` to be whatever you want to call your own branch), then check out that new branch (e.g. `git switch pr26`). Since I enabled maintainer edits on this PR, you should be able to push.

> The BooleanParameterType was part of PR #24

Ah, I had missed that, thanks. I removed the duplicate definition here.

> See https://issues.omg.org/issues/XTCE12-400 for a discussion of the byte ordering changes between XTCE 1.1 and 1.2.

Thanks for the link. That provides some good insight.

Cheers!

medley56:
Awesome! Thanks for the tip on checking out PR branches from forks. I've always struggled with that. 

The last piece here is unit test coverage for these new features. Do you want to tackle that? If not, I can start working on it, adding commits to this branch.

cgobat:
I am pretty tight for time these days. I think it will definitely be ready sooner if you go ahead get started on the tests.

medley56:
Ok. I'll take over this PR branch and start working on tests.

space_packet_parser/xtcedef.py
```diff
@@ -1220,20 +1275,29 @@ def __init__(self, encoding: str = 'utf-8',
             Function that linearly adjusts a size. e.g. if the size reference parameter gives a length in bytes, the
             linear adjuster should multiply by 8 to give the size in bits.
         """
-        if encoding not in ['utf-8', 'utf-16-le', 'utf-16-be']:
-            raise ValueError(
-                f"Got encoding={encoding}. Encoding must be one of utf-8, utf-16-le, or utf-16-be (note that"
-                f"endianness must be specified for utf-16 encoding.")
+        if encoding.upper() not in self._supported_encodings:
```
greglucas, line 1278: This will fail for the `Windows-1252` comparison because it isn't uppercased in the `_supported_encodings`.

The flipping between incoming/upper/lower is a bit confusing through here. Can we not do the upper/lower conversions and rely on the user matching the spec? Or store this as upper-case on the `self.encoding` attribute and then special-case Windows later in the lookup.

image

space_packet_parser/xtcedef.py
```diff
@@ -969,7 +969,27 @@ def read(self, format_string: str, update_position: bool = True) -> Union[int, f
         : Union[int, float, str, bytes]
             Value read from the packet data according to the format specifier.
         """
-        # pylint: disable=too-many-branches
+        # pylint: disable=too-many-branches, too-many-return-statements
+
+        def _twos_complement(val: int, bit_width: int):
```
greglucas, line 974: Were you able to profile this at all? I'm slightly concerned about defining the function repeatedly in here each time through the loop, so it might be beneficial to put it in the outer module scope.

space_packet_parser/xtcedef.py
```diff
@@ -984,13 +1004,33 @@ def read(self, format_string: str, update_position: bool = True) -> Union[int, f
         if update_position:
             self.pos += nbits
 
-        if name == "uint":
+        # TODO: This logic is messy. When we tackle getting away from the bitstring format strings,
+        #  refactor this to be more concise
```
greglucas, line 1008: Yeah, agreed. I'll try and profile it later if you haven't already, but I'd like to make sure that this isn't slowing things up _too_ much with all these added cases to check.

space_packet_parser/xtcedef.py
```diff
@@ -1180,38 +1230,43 @@ def parse_value(self, packet_data: PacketData, parsed_data: dict, **kwargs) -> T
...
 
-        TODO: implement ByteOrderList to inform endianness.
-         This can also relax the requirements on the encoding spec since utf-16-le is redundant if endianness
-         comes from the ByteOrderList
-
         Parameters
         ----------
         encoding : str
-            One of 'utf-8', 'utf-16-le', or 'utf-16-be'. Describes how to read the characters in the string.
+            One of the XTCE-supported encodings, e.g. UTF-8, UTF-16LE.
```
greglucas, line 1252: Maybe inject the encodings here?
```suggestion
            One of the XTCE-supported encodings, {self._supported_encodings}
```

space_packet_parser/xtcedef.py
greglucas: This doesn't seem to make a ton of difference, so no need to fix this up now.

space_packet_parser/xtcedef.py
greglucas: On my local test case (IDEX packet definition), I went from 6.6 to 6.8 seconds. So not miserable, but it did show up. I think this is acceptable as-is to keep moving forward.

space_packet_parser/xtcedef.py
medley56: I just went down a rabbit hole of defining canonicalized enums for all these string parameters, allowing aliases but coercing to standard representations, and just decided it's more complicated than it's worth. I'm OK with demanding strict adherence to the spec.

How do you feel about this when it comes to integer encodings? For example `signed` is not a valid encoding according to XTCE. It must be `twosComplement`.

space_packet_parser/xtcedef.py
medley56: This was throwing pylint and pycodestyle errors but I can suppress them.

PR #43, opened 2024-03-28
MNT: re-use external function in parser loop

greglucas:
# Title of PR

This code block is already identified above, so rely on it there rather than reimplementing the same logic twice.

The only difference is the elif -> else, but I don't think there are any other routes, so the else should save us an isinstance check as well.

## Checklist


medley56 (approved):
Yep.

PR #51, opened 2024-07-30
ENH: Change parsing logic to bytes manipulation directly

greglucas:
# Speed up parsing by removing bitstring reads

Avoid bitstring and do the bitwise reading and manipulations ourselves with bitwise logic. This greatly increases the performance of the parser by ~3-5x in the read routine. This is sort of the MVP for speed-ups by just implementing the routines we need and trying to replicate bitstring. We could do better by refactoring the `_get_format_string()` logic and avoiding any string matching/parsing/splitting too.

I did not fully remove bitstring at this point since there is a lot of logic surrounding it in `parser.py`. We could make it an optional dependency, but it would be nice to get rid of the `legacy_parse_packet()` method if we go that route. Happy to add that deprecation path here if it'd be useful, but probably a major version bump then.

### Profiling

**main**
image

**this branch**
image

## Checklist

### DIFF
--- a/space_packet_parser/parser.py
+++ b/space_packet_parser/parser.py
@@ -122,7 +122,8 @@ class PacketParser:
         for item in CCSDS_HEADER_DEFINITION:
             header[item.name] = ParsedDataItem(name=item.name,
                                                unit=None,
-                                               raw_value=_extract_bits(packet_data, current_bit, item.nbits))
+                                               # pylint: disable=protected-access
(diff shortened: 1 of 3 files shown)


medley56 (approved):
This looks good to me. All the unit and integration tests pass.
I think no matter what, the next version is going to be a major bump, so let's go ahead and remove bitstring entirely, but let's do it in a separate PR. LGTM!

PR #52, opened 2024-07-30
MNT: Use direct operator evaluation rather than eval()

greglucas:
We can get the dunder attribute for comparison which is faster than calling eval on the expression. Also clean up some of the uses of `.values()` in the comparisons.

## Profiling

Goes from 1.25s in my test case to about 0.25s.
**main**
image
**this branch**
image

## Checklist


medley56 (approved):
Awesome. Nice improvement here. eval is gross.

PR #53, opened 2024-07-31
Add a Parseable Protocol and parse method

greglucas:
This enables the same call structure when parsing the data fields and easier pass-through. It is also faster in the hot-loop than using `isinstance()` to get the specific path to go down.

This builds upon the other two performance PRs as I'm whittling things down here and the calling type definitions change, so it was easier to test on top of everything. I can remove them if that would be helpful, but for review you can just look at the final commit here: 98491cec9df52926849aae7930c44b084812ec50

Changes:
* Added a `parse()` method to `SequenceContainer` and `Parameter` with the same call signature.
  * (Note that this is actually the same call signature as `parse_value()`, but with different return types. I'm wondering if there is actually a way to structure these all the same...)
* Moved `ParsedDataItem` into the `xtcedef` module as it is now needed for the `parse()` method. Also had to reimport into the `parser` module for backwards compatibility.
* Added a `Parseable` `Protocol`, not completely necessary, but seems like something nice to define.
* Changed to using dataclasses for the data structures. These automatically bring along the `AttrComparable`-like stuff and this cleaned up some of the init/repr things IMO. Again, not totally necessary, but I think it'd be nice to start removing some of the boilerplate code if we can.

## Checklist

### DIFF
--- a/tests/unit/test_parser.py
+++ b/tests/unit/test_parser.py
@@ -16,8 +16,10 @@ from space_packet_parser import parser
 )
 def test_parsed_data_item(name, raw_value, unit, derived_value, short_description, long_description, valid):
     """Test ParsedDataItem"""
-    if valid:
-        pdi = parser.ParsedDataItem(name, raw_value, unit, derived_value, short_description, long_description)
-    else:
-        with pytest.raises(ValueError):
-            pdi = parser.ParsedDataItem(name, raw_value, unit, derived_value, short_description, long_description)
+    pdi = parser.ParsedDataItem(name, raw_value, unit, derived_value, short_description, long_description)
+    assert pdi.name == name
+    assert pdi.raw_value == raw_value
+    assert pdi.unit == unit
+    assert pdi.derived_value == derived_value
+    assert pdi.short_description == short_description
+    assert pdi.long_description == long_description
--- a/space_packet_parser/parser.py
+++ b/space_packet_parser/parser.py
@@ -83,10 +83,11 @@ class PacketParser:
         header = {}
         current_bit = 0
         for item in CCSDS_HEADER_DEFINITION:
-            header[item.name] = xtcedef.ParsedDataItem(name=item.name,
-                                               unit=None,
-                                               # pylint: disable=protected-access
-                                               raw_value=xtcedef._extract_bits(packet_data, current_bit, item.nbits))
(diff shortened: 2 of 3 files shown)


medley56:
I want to wait on this until the test fixes are merged in so we can ensure that all the tests are actually passing with these changes.

medley56:
I like this addition of the Protocol interface pattern. Before we go ahead and merge it, what do you think about making Protocols for more of the classes? For example, maybe an `Evaluatable` (for `evaluate`), `ValueParser` (for `parse_value`), `XMLDefined` (for `from_xml_element`). 

I also want to understand better the value we're getting out of `dataclass`.

greglucas:
> Before we go ahead and merge it, what do you think about making Protocols for more of the classes? For example, maybe an Evaluatable (for evaluate), ValueParser (for parse_value), XMLDefined (for from_xml_element).

I don't know enough about what other protocols we want and was really wanting to refactor just the `parse()` logic which was taking place both in `parser` and `xtcedef` before. So, I'd prefer to leave those for future updates and additions.

> I also want to understand better the value we're getting out of dataclass.

I think it somewhat simplifies the logic and makes it explicit that this is a data container, but that is somewhat a personal preference here. One thing it does get you is attribute comparisons automatically without needing the attrcomparable superclass you made.

tests/unit/test_parser.py
```diff
@@ -16,8 +16,10 @@
 )
 def test_parsed_data_item(name, raw_value, unit, derived_value, short_description, long_description, valid):
     """Test ParsedDataItem"""
-    if valid:
-        pdi = parser.ParsedDataItem(name, raw_value, unit, derived_value, short_description, long_description)
-    else:
-        with pytest.raises(ValueError):
-            pdi = parser.ParsedDataItem(name, raw_value, unit, derived_value, short_description, long_description)
```
greglucas, line 23: Note that the final commit did remove this ValueError being raised from the class creation.

I profiled the code and the `__postinit__` was being called for every single creation of the items (added about 0.2 seconds to a 6 second run), so it is in the hot loop and just checking whether the name or raw value is None. IMO, this isn't needed and if we create a data item with None in either of those that is OK and just what we got from the parser.

medley56 (approved):

PR #54, opened 2024-08-29
FIX: Enumerated parameter lookup of duplicate labels

greglucas:
If an entry has duplicate labels for different raw values we need to lookup based on values rather than labels. The values are the unique piece of the enumeration, so swap the dictionary key/value storage.

## Checklist


medley56 (approved):
Approved. Don't know why I went with label:value before. This makes much more sense.

Issue #30, opened 2024-03-19
Fix incorrect parsing of `StringDataEncoding` element


medley56:
Oddly, the StringDataEncoding/SizeInBits element does not behave the same as the BinaryDataEncoding/SizeInBits element. This led to some mistakes in the way StringDataEncoding elements are parsed and the way string data is returned.

In particular, known issues:

1. We currently allow StringDataEncoding/SizeInBits/Fixed to contain DiscreteLookup and DynamicValue. This is incorrect as Fixed may only contain the FixedValue element. (p. 72).
2. We do not currently support the StringDataEncoding/Variable element. The Variable element functions similarly to SizeInBits with its TerminationChar and LeadingSize functionality but also allows DiscreteLookup and DynamicValue as well. (p. 73).
3. The way we parse strings is actually incorrect. The spec says that for strings encoded with SizeInBits (as opposed to Variable), the Fixed/FixedValue size is the maximum size of the string. The LeadingSize and TerminationChar elements actually provide a mechanism for deriving a substring from the full string encoding rather than actually specifying a dynamic size. This is actually convenient in many ways and should be handled by returning the full Fixed/FixedValue string as the raw value and then the substring (using LeadingSize or TerminationChar) as a derived value.

medley56:
@cgobat In reading through the green book I realized there are some fundamental issues with the way we use StringDataEncoding. I think these issues are mostly separate from the work you've done with #26 but I wanted to let you know.

cgobat:
Copy that&mdash;thanks for the heads up.

PR #58, opened 2024-08-30
Fix Unit Tests Relying on ConstBitStream to use PacketData Instead

medley56:
# Fix Unit Tests Relying on ConstBitStream to use PacketData Instead

These tests were passing because the input data was a ConstBitStream which is still handled (until we rip it all out) so we weren't really testing the PacketData object's ability to parse data. 

- Update StringDataEncoding tests to use PacketData instead of ConstBitStream
- Add some more type-hinting throughout xtcedef.py
- Update unit tests for integer parsing using PacketData instead of ConstBitStream 
- Update float encoding unit tests to use PacketData 
- Update enumerated parameter type tests to use PacketData 
- Update binary parameter parsing test to use PacketData 
- Update boolean parameter parsing tests to use PacketData 
- Update time parameter type parsing tests to use PacketData instead of ConstBitStream

## Checklist
- [ ] The changelog.md has been updated - I don't think we need to update the changelog for this.

### DIFF
--- a/space_packet_parser/xtcedef.py
+++ b/space_packet_parser/xtcedef.py
@@ -4,9 +4,8 @@ from abc import ABCMeta
 from collections import namedtuple
 import inspect
 import logging
-from pathlib import Path
 import struct
-from typing import Tuple, Union, Optional, Any, List
+from typing import Tuple, Union, Optional, Any, List, TextIO, Dict
 import warnings
 from xml.etree import ElementTree
 
@@ -159,7 +158,7 @@ class Comparison(MatchCriteria):
                              f"{set(self._valid_operators.keys())}")
 
     @classmethod
-    def from_match_criteria_xml_element(cls, element: ElementTree.Element, ns: dict):
+    def from_match_criteria_xml_element(cls, element: ElementTree.Element, ns: dict) -> 'Comparison':
         """Create
 
         Parameters
@@ -171,7 +170,7 @@ class Comparison(MatchCriteria):
 
         Returns
         -------
-        : cls
+        : Comparison
         """
         use_calibrated_value = True  # Default
         if 'useCalibratedValue' in element.attrib:
@@ -245,8 +244,13 @@ class Condition(MatchCriteria):
     but it's functionally close enough that we inherit the class here.
     """
 
-    def __init__(self, left_param: str, operator: str, right_param: Optional[str] = None, right_value: Optional[Any] = None,
-                 left_use_calibrated_value: bool = True, right_use_calibrated_value: bool = True):
+    def __init__(self,
+                 left_param: str,
+                 operator: str,
+                 right_param: Optional[str] = None,
+                 right_value: Optional[Any] = None,
+                 left_use_calibrated_value: bool = True,
+                 right_use_calibrated_value: bool = True):
         """Constructor
 
         Parameters
@@ -404,7 +408,7 @@ class BooleanExpression(MatchCriteria):
         self.expression = expression
 
     @classmethod
-    def from_match_criteria_xml_element(cls, element: ElementTree.Element, ns: dict):
+    def from_match_criteria_xml_element(cls, element: ElementTree.Element, ns: dict) -> 'BooleanExpression':
         """Abstract classmethod to create a match criteria object from an XML element.
 
         Parameters
@@ -416,10 +420,10 @@ class BooleanExpression(MatchCriteria):
 
         Returns
         -------
-        : cls
+        : BooleanExpression
         """
 
-        def _parse_anded(anded_el: ElementTree.Element):
+        def _parse_anded(anded_el: ElementTree.Element) -> Anded:
             """Create an Anded object from an xtce:ANDedConditions element
 
             Parameters
@@ -436,7 +440,7 @@ class BooleanExpression(MatchCriteria):
             anded_ors = [_parse_ored(anded_or) for anded_or in anded_el.findall('xtce:ORedConditions', ns)]
             return Anded(conditions, anded_ors)
 
-        def _parse_ored(ored_el: ElementTree.Element):
+        def _parse_ored(ored_el: ElementTree.Element) -> Ored:
             """Create an Ored object from an xtce:ARedConditions element
 
             Parameters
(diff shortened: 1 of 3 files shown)


space_packet_parser/xtcedef.py
```diff
@@ -936,15 +940,41 @@ def __init__(self, data: bytes):
...
+            Whether to update the cursor position in the packet. Default True.
+
+        Returns
+        -------
+        : Union[int, float, str, bytes]
+            Value read from the packet data according to the format specifier.
+        """
+        # pylint: disable=too-many-branches
+        name, n_things = format_string.split(":")
+        if name == "bytes":
```
medley56, line 973: As we discussed.

space_packet_parser/xtcedef.py
```diff
@@ -956,13 +986,17 @@ def read(self, format_string, update_position=True):
         if name == "uint":
             return bytes_as_int
         if name == "int":
-            return int.from_bytes(int.to_bytes(bytes_as_int, nbits // 8 + 1, byteorder="big"),
-                                  byteorder="big", signed=True)
+            # Compute two's complement for signed integer of any size (nbits)
+            if (bytes_as_int & (1 << (nbits - 1))) != 0:  # if sign bit is set e.g., 8bit: 128-255
+                return bytes_as_int - (1 << nbits)  # compute negative value
+            return bytes_as_int  # return positive value as is
         if name == "floatbe":
```
medley56, line 993: Had to update the struct format strings in here to force big endianness.

space_packet_parser/xtcedef.py
```diff
@@ -1243,14 +1300,16 @@ def _get_format_string(self, packet_data: PacketData, parsed_data: dict):
             bits_per_byte = 8
             look_ahead_n_bytes = 0
             while look_ahead_n_bytes <= len(packet_data) - packet_data.pos:
-                look_ahead = packet_data.peek(f'bytes:{look_ahead_n_bytes}')  # Outputs UTF-8 encoded byte string
-                look_ahead = look_ahead.decode('utf-8').encode(self.encoding)  # Force specified encoding
```
medley56, line 1247: I don't know why I was decoding in utf8 and then re-encoding in whatever the specified encoding was (e.g. utf-16-be). All we need to do in order to find the termination character is compare literal byte strings, regardless of how they are encoded.

greglucas (approved):

tests/unit/test_xtcedef.py
```diff
@@ -1148,7 +1147,7 @@ def test_string_parameter_type(xml_string: str, expectation):
                                        use_calibrated_value=False,
                                        length_linear_adjuster=lambda x: 8*x)),
          {'STR_LEN': parser.ParsedDataItem('STR_LEN', 8, None)},
-         '0b01000010010000010100010000100000010101110100111101001100010001100100011101000001010100100100001001000001',
+         xtcedef.PacketData(b'BAD WOLF'),
```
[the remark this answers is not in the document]
medley56: Well, not _all_, but most are better. There are still a few edge cases that I did with literal binary strings (ints).

space_packet_parser/xtcedef.py
medley56: I've change the look ahead to look 16 characters at a time rather than one character at a time. This means ~1/16 as many peek operations, and for many strings, we'll catch it in the first peek. And I am splitting the peeked bytes on the termination character now, as you suggested.

greglucas: This is OK, but I guess my more general comment/question was whether we can read the entire string data chunk at once and do this check, rather than the while-loop peeking.

Here is some nice documentation about the various cases:
https://docs.yamcs.org/yamcs-server-manual/mdb/data-types/#string-data-type

I _think_ that my thought is we'd just read the entire "string-buffer" (fixed size of this parameter, or until the end of the packet). That would be our maximum length string. Then the termination character is just used to trim that potential string I think? (i.e I wrote my packet definition to be 8 bytes for this spot, but for this downlink I am only sending `b"HELLO"` and filling 5 of those, with `b"\x00\x00\x00"` that we want to trim off the end and not show to the user.

So, we would always go and get all the bytes for the full string-buffer. Then split on termination character and return the first occurrence. (It also looks like it is acceptable to not terminate the string if the string just ends at the end of the buffer)

The issue here might be that this logic actually needs to be in the `parse_value()` method instead of here. In fact, it might currently be incorrect and setting the position of the data packet to after the termination character rather than after the fixed size field. I don't know enough about all of this, so just putting thoughts down here in case it sparks anything on your end.

greglucas (approved):

space_packet_parser/xtcedef.py
medley56: There are many ways to specify a string length, a termination character is only one. I think if you specify a termination character it has to be there (it can't just end with the end of the buffer). We could peek the entire buffer to find that termination character but that could be 500mb of data if we've read an entire packet file. The basic problem is we have no idea how long the string is. But maybe a compromise could be peeking either the max size of a CCSDS packet from where we currently are, or the full remaining buffer, whichever is shorter. Figure out the string length from that, and send it back.

You're correct that some of this happens in parse_value. That's why this method sends back a skip_bits_after integer. It tells parse_value to skip the termination character after reading the string.

greglucas: In here, we only ever have PacketData being passed, not the entire file. We are always parsing individual packets and the bytes associated with that one piece, so we are helped by that and will only ever read to the end of the packet length.

I just found this: https://github.com/medley56/space_packet_parser/issues/30
and I think your (3) bullet is the issue I think I'm bringing up.

Regardless, this can be considered for follow-up and a future fix and I think we should get this in. I was just bringing it up because it caught my eye here.

medley56: Ohhh, right! I had forgotten about that. That seems like a nice small ticket to handle next without getting in the way of the bitstring removal and CSV deprecation work. Let's merge this as is, because it does match the current behavior and this kind of logic will still be necessary for `Variable` length defined strings, but this is not its final form.

PR #59, opened 2024-08-31
Remove bitstring dependency

greglucas:
# Title of PR

This removes the bitstring dependency and all bitstring inputs allowed with it. You can pass in the raw objects instead of the bitstring objects now.

~~This also removes the `legacy_parser` method which was kept around to support csv parsing. This has been removed in favor of converting the csv packet definitions into the XTCE format instead.~~

This sits on top of https://github.com/medley56/space_packet_parser/pull/58 so there wasn't as much thrashing going on. Only the second commit is the removal portion.

~~NOTE: I have removed the entire csvdef module. I am not sure if you want to keep that around or not. This is definitely the heavy-handed approach of just removing all of the legacy things and moving on in a major version bump.~~

## Checklist

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -6,9 +6,11 @@ list and release milestones.
 Release notes for the `space_packet_parser` library
 
 ### v5.0.0 (unreleased)
-- Replace bitstring objects with native Python bytes objects
-  - Remove dependency on the bitstring library
+- BREAKING: Replace `bitstring` objects with native Python bytes objects
+  - Remove dependency on the `bitstring` library
   - Much faster parsing speed
+  - Users that are passing `bitstring.ConstBitStream` objects to `generator` will need to pass a 
+    binary filelike object instead
 - Fix EnumeratedParameterType to handle duplicate labels
 - Add error reporting for unsupported and invalid parameter types
 
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -6,7 +6,9 @@ list and release milestones.
 Release notes for the `space_packet_parser` library
 
 ### v5.0.0 (unreleased)
-- Replace bitstring objects with native Python bytes objects (except for in CSV-based parsing code)
+- Replace bitstring objects with native Python bytes objects
+  - Remove dependency on the bitstring library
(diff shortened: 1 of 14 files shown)


medley56:
Looks like there are some issues with the tests hanging around still. Some of the integration tests for CSV parsing we will either need to remove entirely or we will need to come up with an XTCE definition for the previously used CSV files.

greglucas:
> Looks like there are some issues with the tests hanging around still. Some of the integration tests for CSV parsing we will either need to remove entirely or we will need to come up with an XTCE definition for the previously used CSV files.

For some reason, I thought the csv parsing required bitstring which is why I removed it earlier. It turns out it doesn't, so I've now reverted all of the CSV changes and made this just a removal of bitstring now. This is much cleaner and can take place in two steps with a future removal of the legacy csv parser.

docs/source/changelog.md
```diff
@@ -6,7 +6,9 @@ list and release milestones.
 Release notes for the `space_packet_parser` library
 
 ### v5.0.0 (unreleased)
-- Replace bitstring objects with native Python bytes objects (except for in CSV-based parsing code)
+- Replace bitstring objects with native Python bytes objects
```
medley56, line 9: Leave a note in here that this is the breaking change and anyone using ConstBitStream objects as input to the generator will need to change to the (much prettier) direct input of a binary filelike object.

medley56 (approved):
This is awesome. Unless my recollection deceives me, the speed boost is perceptible in the integration tests.

docs/source/changelog.md
```diff
@@ -6,7 +6,9 @@ list and release milestones.
...
+  - Remove dependency on the bitstring library
+  - Much faster parsing speed
```
medley56, line 11:
```suggestion
- BREAKING: Replace `bitstring` objects with native Python bytes objects
  - Remove dependency on the `bitstring` library
  - Much faster parsing speed
  - Users that are passing `bitstring.ConstBitStream` objects to `generator` will need to pass a 
    binary filelike object instead
```

medley56 (approved):

PR #61, opened 2024-09-01
FIX: Return the raw bytes from a BinaryDataEncoding

greglucas:
When dealing with binary data, users generally want the most raw data which is the bytes directly rather than a string representation of the bits. This is a breaking change and if users want to get the binary string, they can format the string output themselves now.

# Title of PR

Brief description of changes. If you write good commit messages, put the concatenated list of messages here.

## Checklist

### DIFF
--- a/tests/unit/test_xtcedef.py
+++ b/tests/unit/test_xtcedef.py
@@ -1704,7 +1704,7 @@ def test_binary_parameter_type(xml_string: str, expectation):
             xtcedef.BinaryDataEncoding(fixed_size_in_bits=16)),
          {},
          xtcedef.PacketData(0b0011010000110010010100110000000001001011000000000100100100000000.to_bytes(length=8, byteorder='big')),
-         '0011010000110010'),
+         b"42"),
         # discrete lookup list size
         (xtcedef.BinaryParameterType(
             'TEST_BIN',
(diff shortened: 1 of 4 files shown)


tests/unit/test_xtcedef.py
```diff
@@ -1762,7 +1762,7 @@ def test_boolean_parameter_type(xml_string, expectation):
             xtcedef.BinaryDataEncoding(fixed_size_in_bits=1)),
          {},
          xtcedef.PacketData(0b0011010000110010010100110000000001001011000000000100100100000000.to_bytes(length=64, byteorder='big')),
-         '0', True),
+         b'', False),
```
greglucas, line 1765: I think this was actually incorrect before. If I am reading in the 0 bit as a boolean data type I think that should be False. Before we were getting a string representation of the bits which is `"0"`, and `bool("0") == True` because there is something inside the string and it isn't empty.

medley56: I agree. 

I struggled a bit with truthiness when writing tests. You'll see tests for string encoded data types with values like "false_is_truthy" when encoding a boolean type as a string. Basically unless the value is empty, it became True and I eventually convinced myself that, for strings at least, that's what I wanted. And since I was outputting binary data as strings, the behavior stuck here too.

However, since binary data is basically an int, anything that's all zeros should be falsy. Everything else should be truthy (nonzero). This is a change in behavior but not one that anyone was relying on I'm pretty sure.

medley56 (approved):
You have an extra return in the changelog but other than that, this looks good.

medley56 (approved):

PR #62, opened 2024-09-02
MNT: Simplify the string parsing termination character handling

greglucas:
This does a string index lookup for the termination character instead of needing to track everything through the while loop and hitting the `packet_data.peek()` method multiple times. This is a refactor and doesn't address the other items of string parsing needing to be updated.

Note, I'm still a bit unclear about all of the various cases for handling termination character even though I've read the spec a few times now... Some more full examples would definitely help. Going off of this example: https://github.com/yamcs/yamcs/blame/78b9553caf3c9f7ef7a6e6897d236a69aeed8190/yamcs-core/src/test/resources/xtce/strings-tm.xml#L37-L53
I'm actually wondering if when using the termination character as the delimiter you still need to include a maxSizeBits value through the Variable Element?

xref: https://github.com/medley56/space_packet_parser/issues/30

## Checklist

### DIFF
--- a/space_packet_parser/xtcedef.py
+++ b/space_packet_parser/xtcedef.py
@@ -1308,15 +1308,6 @@ class StringDataEncoding(DataEncoding):
             raise ValueError(f"Got encoding={encoding} (uppercased). "
                              f"Encoding must be one of {self._supported_encodings}.")
         self.encoding = encoding
-        # Check that the termination character is a single character in the specified encoding
-        # e.g. b'\x58' in utf-8 is "X"
-        # b'\x21\00' in utf-16-le is "!"
-        # b'\x00\x21' in utf-16-be is "!"
-        if termination_character and len(bytes.fromhex(termination_character).decode(encoding)) != 1:
-            raise ValueError(f"Termination character {termination_character} appears to be malformed. Expected a "
-                             f"hex string representation of a single character, e.g. '58' for character 'X' in UTF-8 "
-                             f"or '5800' for character 'X' in UTF-16LE. Note that variable-width encoding is not "
-                             f"yet supported in any encoding.")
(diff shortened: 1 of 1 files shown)


space_packet_parser/xtcedef.py
medley56, line 1: Return docstring for `StringDataEncoding._get_format_string` needs updating.

space_packet_parser/xtcedef.py
```diff
@@ -1379,58 +1383,23 @@ def _get_format_string(self, packet_data: PacketData, parsed_data: dict) -> Tupl
...
-                # Ensure we never go over the total length of the packet, in bits
-                look_ahead_n_bytes = min(look_ahead_n_bytes, n_full_bytes_left_in_packet)
-            else:
-                raise ValueError(f"Reached end of binary string without finding "
-                                 f"termination character {self.termination_character}.")
+            # Look through the rest of the packet data to find the termination character
+            nbits_left = len(packet_data) - packet_data.pos
+            string_buffer = packet_data.peek(f"bytes:{nbits_left//8}")
+            try:
+                strlen_bits = string_buffer.index(self.termination_character) * 8
```
medley56, line 1390:
```suggestion
                strlen_bits = string_buffer.index(bytes.fromhex(self.termination_character)) * 8
```

greglucas: This is one of the places where I put the `bytes.fromhex()` call up into the `__init__` function so it is just called once per packet type rather than each time we read a packet. (`self.termination_character` is the bytes representation at this point)

medley56: That works for me.

medley56 (approved):
This is great! Thanks! I also like the change that skips the termination character during parsing rather than returning a wonky additional value from the format string method. Cleans things up nicely!

PR #63, opened 2024-09-03
Faster value parsing

### LINKED ISSUE #57: Move Away from Bitstring Format Strings
In all of our data encoding classes (e.g. `StringDataEncoding`), we have a standard interface method called `_get_format_string`, which returns a bitstring-style format string like "bytes:5" (for 5 bytes) or "uint:12" (for a 12-bit unsigned int). These were very convenient when all of our parsing was based on the bitstring library but now that we are using Python bytes objects directly, we can be more efficient about how we communicate what to parse.

Our first idea is to pass a tuple with a type and a length. e.g. instead of `"bytes:5"`, pass `("bytes", 5)`. This saves a string split step inside the `PacketData.parse` method, which would otherwise be called for every data item in every packet.

Note that `StringDataEncoding._get_format_string` actually returns an additional value for the number of bits to skip after parsing. This is to support the concept of a string that is terminated with a termination character or string such as `\x00\x00\xff\xff`, which shouldn't be part of the parsed value but instead should be skipped over after parsing. This position update (`packet_data.pos += skip_bits_after`) is actually handled inside of `StringDataEncoding`, so the interface to `PacketData.parse` is unchanged but its internal `pos` attribute _is_ being modified from an external source.

greglucas:
This is a proof of concept PR for now, building on top of the little-endian PR to reduce the thrashing when rebasing, so only the last commit is relevant. It is currently failing one test because of the binary-bytes transition that would make this PR easier as well.

I get 5.2s with this PR, vs 6.6s from main, so a pretty significant speed-up.

The major changes are:
- Move some parameter instantiation logic up into the init routines rather than in the hot-loop reading packets.
- Make a `read_bytes` and `read_int` method on `PacketData` for the two explicit cases of wanting to get data without having to convert between the two so frequently.
- Move the if-branches from the `read()` method up into their respective data encodings so that we limit the number of checks needed when returning the parsed data.

### TODO

~~Remove `PacketData.read()` / `PacketData.peek()` methods once the string refactor work goes in.~~

**Question:**
Do we want to make the `DataEncoding` parsing a standard workflow? Something like the following:
```python
def parse_value(...):
    self._calculate_size()
    self._get_raw_value()
    self._calibrate_value()
```

where each subclass could override those specific functions rather than `parse_value()` directly. For example, `StringDataEncoding` wouldn't need a `calibrate_value()` because there are no calibrators, so it could be a no-op there. Currently, we are doing something similar with `get_bitstring_format()`, but there are differing amounts of logic in each subclass location and wondering if it'd make sense to expand into a few different routines instead.

closes #57

## Checklist

### DIFF
--- a/tests/unit/test_xtcedef.py
+++ b/tests/unit/test_xtcedef.py
@@ -1523,7 +1523,7 @@ def test_float_parameter_type(xml_string: str, expectation):
             'TEST_FLOAT',
             xtcedef.FloatDataEncoding(32, byte_order='leastSignificantByteFirst')),
          {},
-         xtcedef.PacketData(0b01000000010010010000111111010000.to_bytes(length=4, byteorder='big')),
+         xtcedef.PacketData(0b01000000010010010000111111010000.to_bytes(length=4, byteorder='little')),
          3.14159),
         # Test big endian 64-bit float
         (xtcedef.FloatParameterType('TEST_FLOAT', xtcedef.FloatDataEncoding(64)),
@@ -1704,7 +1704,7 @@ def test_binary_parameter_type(xml_string: str, expectation):
             xtcedef.BinaryDataEncoding(fixed_size_in_bits=16)),
          {},
          xtcedef.PacketData(0b0011010000110010010100110000000001001011000000000100100100000000.to_bytes(length=8, byteorder='big')),
-         b"42"),
+         b'42'),
         # discrete lookup list size
         (xtcedef.BinaryParameterType(
             'TEST_BIN',
@@ -1716,7 +1716,7 @@ def test_binary_parameter_type(xml_string: str, expectation):
             ], linear_adjuster=lambda x: 8*x)),
          {'P1': parser.ParsedDataItem('P1', 1, None, 7.4)},
          xtcedef.PacketData(0b0011010000110010010100110000000001001011000000000100100100000000.to_bytes(length=8, byteorder='big')),
-         b"42"),
+         b'42'),
         # dynamic size reference to other parameter
         (xtcedef.BinaryParameterType(
             'TEST_BIN',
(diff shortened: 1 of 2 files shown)


tests/unit/test_xtcedef.py
```diff
@@ -1796,7 +1796,7 @@ def test_boolean_parameter_type(xml_string, expectation):
             xtcedef.BinaryDataEncoding(fixed_size_in_bits=1)),
          {},
          xtcedef.PacketData(0b0011010000110010010100110000000001001011000000000100100100000000.to_bytes(length=64, byteorder='big')),
-         b'', False),
+         b'\x00', True),
```
greglucas, line 1799: I think I messed up previously 🐑 ,
https://github.com/medley56/space_packet_parser/commit/15af144257dda66744bbcbc9595d0dd0107dc495#diff-56c0e979a3995704b94f150fb764e7e524b03a092ad029050c1d99c6c6115cd2R1799

I'm now going back on what I said the last time and think this should actually be as you had it before. We are saying this is `BinaryDataEncoding`, which means we want bytes back. The last update I made squeezed the data down into the smallest number of bytes. Whereas with this change we are padding with 0s and expanding to fit the data within a byte.

You are emitting a warning already telling people this is a bad idea.

medley56: Hmm, agree about padding for sure. It still seems like a binary string of zeros should probably be False, but we're just defaulting to Python behavior, so they've probably thought about it more than I have in `bool(b'\x00')` is really evaluating whether there is any data in the byte string. I guess the way to actually check it's value would need to be at the bit level, like `bool(int.from_bytes(b'\x00', byteorder='big') & 0xFF)`.

All that to say, let's go with the Python behavior for evaluating the truthiness of binary strings. The XTCE spec even says on the subject of BooleanParameterType, "Binary encoding is typically a 'pass through' raw encoding form...".

medley56:
This approach makes a lot of sense. The logic for parsing a value is tightly coupled to the DataEncoding and I don't think there's much risk of duplicating parsing logic between encoding types because, well, they are encoded differently. 

Let me know when this is beyond a POC.

greglucas:
@medley56 
>Let me know when this is beyond a POC.
I think this is good to go now as long as you are OK with it as-is. If you see some `Protocol`s or `ABC`s that would make sense here we can try to add them later too.

medley56 (approved):
LGTM. I want this in before I start incorporating my MIL 1750 parsing.

PR #67, opened 2024-09-04
Unify packet class

greglucas:
This combines the `PacketData` and `Packet` classes together. A `Packet` now contains all of the raw bytes, and can be "parsed" as it goes keeping track of the reading. This enables us to avoid having to pass in `parsed_data` to the parsing methods and just pass in the `Packet` container that already has that information. It would be good to look at the definition of that class and let me know what you think it should/should not contain as public/private methods (maybe we don't want to make the reads public?).

This simplifies some of the tests so we don't always have to pass in `parsed_data={}` now.

This seems to speed up the test suite pretty significantly too. I went from 15s to 7.5 s locally.

## Checklist


medley56 (approved):
Yep, sounds good to me! These interface changes and simplifications are awesome!

PR #68, opened 2024-09-04
Refactor modules

### LINKED ISSUE #34: Refactor `xtcedef.py` into sub modules
The xtcedef.py module is about 2500 lines of code and the accompanying test module is about 1800. These are too big and make navigating the codebase annoying. 

Refactor xtcedef and its test module to split up into different groups. Perhaps:
- common classes (e.g. ABCs)
- exceptions
- encodings
- parameter_types
- parameters

Could put the top level XtceDefinition class either in __init__.py or in its own submodule.

greglucas:
This is a major refactor, breaking the xtcedef.py module into separate entities.

I went with:
* calibrators
* encodings
* exceptions
* matches (comparisons? other ideas for this?)
* packets (contains the sequence container and packet data definitions, would "containers" be a better module name?)
* parameters (contains the `Parameter` (only one) and `ParameterTypes`)

I'm happy to move things around more/less too, just let me know your preferences. There is some fighting going on with circular imports depending on how we structure this, so we do need to make sure not to couple the modules too closely.

closes #34

## Checklist


medley56:
I think I borked this PR by merging your Packet/PacketData unification PR. Sorry about that. This is going to be a pain if we're changing code inside the modules. After my MIL-1750A PR I'm done for a while. Can we merge that, rebase this, and then merge it?

medley56:
@greglucas I'll approve this as soon as you can resolve the rebase.

greglucas:
@medley56, this should be good to go with a rebase now and I can do large find/replace to move things around pretty easily if we want to change any namespaces.

My two questions for name changes (better suggestions welcome too!):
* packets.py -> containers.py ?
* matches.py -> comparisons.py ? (matches was slightly annoying because you used matches as a variable name and so we were shadowing the imported module name. There is also a new match syntax in Python 3.10+ that we might want to try to avoid confusion with)

medley56:
@greglucas I like `comparisons.py` over `matches.py`. I think I like `parsing.py` or maybe `parseables.py` more than `containers.py` because that module contains the Parsable protocol and everything in there seems to be related to actually parsing data.

medley56 (approved):

space_packet_parser/encodings.py
```diff
@@ -0,0 +1,796 @@
...
+        context_calibrators = cls.get_context_calibrators(element, ns)
+        return cls(size_in_bits=size_in_bits, encoding=encoding, byte_order=byte_order,
+                   default_calibrator=calibrator, context_calibrators=context_calibrators)
+
+
+class FloatDataEncoding(NumericDataEncoding):
+    """<xtce:FloatDataEncoding>"""
+    _supported_encodings = ['IEEE-754', 'MIL-1750A']
+
+    def __init__(self, size_in_bits: int, encoding: str = 'IEEE-754',
```
cgobat, line 553: I believe these string restrictions are incorrect. They should be 'IEEE754' and 'MILSTD_1750A', rather than 'IEEE-754' and 'MIL-1750A'. Here's the enumerated value list for `FloatEncodingType` from lines 3886-3895 of the [XTCE schema document](https://www.omg.org/spec/XTCE/20180204/SpaceSystem.xsd):

```xml
<simpleType name="FloatEncodingType">
    <restriction base="string">
        <enumeration value="IEEE754_1985"/>
        <enumeration value="IEEE754"/>
        <enumeration value="MILSTD_1750A"/>
        <enumeration value="DEC"/>
        <enumeration value="IBM"/>
        <enumeration value="TI"/>
    </restriction>
</simpleType>
```

The above also agrees with the list in section 4.3.2.2.5.7.1 of the [element description document](https://public.ccsds.org/Pubs/660x1g2.pdf).

PR #69, opened 2024-09-04
Support MIL-1750A Float Encoding

medley56:
# Support MIL-1750A Float Encoding

Bump python requirement to 3.9 and remove 3.8 tests and github workflow
Add support for MIL-1750A float parsing
Add test cases pulled from MIL-1750A standard document

## Checklist

### DIFF
--- a/space_packet_parser/xtcedef.py
+++ b/space_packet_parser/xtcedef.py
@@ -1,6 +1,6 @@
 """Module for parsing XTCE xml files to specify packet format"""
 # Standard
-from abc import ABCMeta
+from abc import ABCMeta, abstractmethod
 from collections import namedtuple
 from dataclasses import dataclass, field
 import inspect
@@ -111,6 +111,7 @@ class MatchCriteria(AttrComparable, metaclass=ABCMeta):
     }
 
     @classmethod
+    @abstractmethod
     def from_match_criteria_xml_element(cls, element: ElementTree.Element, ns: dict):
         """Abstract classmethod to create a match criteria object from an XML element.
 
@@ -127,6 +128,7 @@ class MatchCriteria(AttrComparable, metaclass=ABCMeta):
         """
         raise NotImplementedError()
 
+    @abstractmethod
     def evaluate(self, parsed_data: dict, current_parsed_value: Optional[Union[int, float]] = None) -> bool:
         """Evaluate match criteria down to a boolean.
 
@@ -609,6 +611,7 @@ class Calibrator(AttrComparable, metaclass=ABCMeta):
     """Abstract base class for XTCE calibrators"""
 
     @classmethod
+    @abstractmethod
     def from_calibrator_xml_element(cls, element: ElementTree.Element, ns: dict) -> 'Calibrator':
         """Abstract classmethod to create a default_calibrator object from an XML element.
 
@@ -625,6 +628,7 @@ class Calibrator(AttrComparable, metaclass=ABCMeta):
         """
         return NotImplemented
 
+    @abstractmethod
     def calibrate(self, uncalibrated_value: Union[int, float]) -> Union[int, float]:
         """Takes an integer-encoded or float-encoded value and returns a calibrated version.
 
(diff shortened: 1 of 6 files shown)


space_packet_parser/xtcedef.py
```diff
@@ -1472,6 +1478,16 @@ def _get_raw_value(self, packet: Packet) -> Union[int, float]:
         """
         raise NotImplementedError()
 
+    @staticmethod
+    def twos_complement(val: int, bit_width: int) -> int:
```
greglucas, line 1482: This seems like a "helper" function to me, kind of like `_extract_bits()`. Do we want to (a) make it private, and (b) just add it to the bottom of the file as an extra module-level function rather than sticking it onto a class?

space_packet_parser/xtcedef.py
```diff
@@ -1581,33 +1593,69 @@ def __init__(self, size_in_bits: int, encoding: str = 'IEEE-754',
...
+                """Parsing function for MIL-1750A floats"""
+                # MIL 1750A floats are always 32 bit
+                # See: https://www.xgc-tek.com/manuals/mil-std-1750a/c191.html#AEN324
+                #
+                #  MSB                                         LSB MSB          LSB
+                # ------------------------------------------------------------------
+                # | S|                   Mantissa                 |    Exponent    |
+                # ------------------------------------------------------------------
+                #   0  1                                        23 24            31
+                bytes_as_int = int.from_bytes(mil_bytes, byteorder='big')
```
greglucas, line 1616: Is MIL always big endian?

space_packet_parser/xtcedef.py
```diff
@@ -1581,33 +1593,69 @@ def __init__(self, size_in_bits: int, encoding: str = 'IEEE-754',
...
+                exponent = bytes_as_int & 0xFF  # last 8 bits
+                mantissa = (bytes_as_int >> 8) & 0xFFFFFFFF  # bits 0 through 23 (24 bits)
```
greglucas, line 1618: Don't you only want this to be 6 Fs if you want the first 24 bits?

tests/unit/test_xtcedef.py
```diff
@@ -1523,16 +1524,73 @@ def test_float_parameter_type(xml_string: str, expectation):
...
+            'MIL_1750A_FLOAT',
+            xtcedef.FloatDataEncoding(32, encoding="MIL-1750A")),
+         xtcedef.Packet(b'\x9F\xFF\xFF\x04'),
+         -0.7500001 * (2 ** 4)),
     ]
 )
 def test_float_parameter_parsing(parameter_type, packet, expected):
     """Test parsing float parameters"""
     raw, derived = parameter_type.parse_value(packet)
+    # NOTE: These results are compared with a relative tolerance due to the imprecise storage of floats
```
greglucas, line 1589: You can avoid the numpy import and use `pytest.approx`: https://docs.pytest.org/en/latest/reference/reference.html#pytest.approx

```python
    expected = pytest.approx(expected, rel=0.000001)
    if derived:
        assert derived == expected
    else:
        assert raw == expected
```

greglucas (approved):
Looks great to me. I've just got a few minor suggestions. We should decide what we want private vs public interfaces.

tests/unit/test_xtcedef.py
medley56: Oh nice! I didn't know about this. Will do.

space_packet_parser/xtcedef.py
medley56: Huh... Yeah. I guess this actually doesn't matter because when we bit shift `bytes_as_int`, it gets padded with zeros. I don't suppose a bitwise AND takes much time but I could just take it out. We do need the mask for the exponent.

space_packet_parser/xtcedef.py
medley56: It's not in the standard, so I'll check the `byte_order`.

space_packet_parser/xtcedef.py
medley56: Agreed. It's only used in our numeric data encodings so I left it on the class. I usually leave static helper methods as public because they can't hurt anything related to the associated class instances but if it's not used elsewhere I can see why it's cleaner to make it private.

medley56:
@greglucas Updated and ready for final review.

greglucas (approved):

PR #77, opened 2024-09-25
MNT: Turn some position arguments into keyword-only

### LINKED ISSUE #76: New pylint release causes failures
The new release of pylint causes some failures with the number of positional arguments allowed. We should either ignore this error, or change some of the parameters to keyword only.

greglucas:
This makes some parameters keyword only as an alternative to skipping the linter altogether. If we go down this path, we should probably make some other parameters keyword only too and not just the ones to satisfy the linter.

closes #76 

## Checklist

### DIFF
--- a/space_packet_parser/comparisons.py
+++ b/space_packet_parser/comparisons.py
@@ -215,6 +215,7 @@ class Condition(MatchCriteria):
     def __init__(self,
                  left_param: str,
                  operator: str,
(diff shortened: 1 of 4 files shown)


medley56:
This was the approach I was going to take actually. Do you think you could take a stab at making the other modules consistent with this (to the extent that you think makes sense)?

PR #79, opened 2024-09-25
Refactor API and Move Parser Methods to XtcePacketDefinition Class

medley56:
# Main API Change

This PR is based on #78 

Users now only need to instantiate a single object from a packet definition document. The parsing methods are directly available on the XtcePacketDefinition class and the parser.py module has been removed.

## Checklist


space_packet_parser/definitions.py
```diff
@@ -385,269 +359,326 @@ def _get_container_base_container(
...
+            n_packets_parsed += 1
+            n_bytes_parsed += skip_header_bytes + n_bytes_packet
+            if ccsds_headers_only is True:
+                # update the current position to the end of the packet data
+                current_pos += n_bytes_packet
+                p = parseables.CCSDSPacket(raw_data=read_buffer[current_pos - n_bytes_packet:current_pos], **header)
+                yield p
+                continue
+
+            # Based on PKT_LEN fill buffer enough to read a full packet
```
greglucas, line 637: Thinking about the logic here, should we actually move this while-loop above the header yield/continue? In that yield statement above we are skipping forward `n_bytes_packet`, but we don't know we have that many bytes available in the buffer yet. So we could be advancing past the end of the read_buffer without realizing it. If we move this read above that I think we would fix that issue and shouldn't really slow anything down because we'd have to read that data anyways...

space_packet_parser/definitions.py
```diff
@@ -385,269 +359,326 @@ def _get_container_base_container(
...
+                                                word_size=self.word_size)
+            except UnrecognizedPacketTypeError as e:
+                logger.debug(f"Unrecognized error on packet with APID {header['PKT_APID'].raw_value}'")
+                if yield_unrecognized_packet_errors is True:
+                    # Yield the caught exception without raising it (raising ends generator)
+                    yield e
+                # Continue to next packet
+                continue
+
+            if packet.header['PKT_LEN'].raw_value != header['PKT_LEN'].raw_value:
```
greglucas, line 662: I was actually wondering about this check and whether we could avoid reading the first 7 items in the container definition and rely on the CCSDS header being correct. That way we could start the Packet class right away when we are doing the header above and then continue on here. Currently we are parsing the header twice essentially.

Could we check this condition in the definition itself somehow so that when someone is creating the definition from the XTCE we would raise immediately before parsing even began?

space_packet_parser/definitions.py
```diff
@@ -385,269 +359,326 @@ def _get_container_base_container(
...
+
+            actual_length_parsed = packet.raw_data.pos // 8
+            if actual_length_parsed != n_bytes_packet:
+                logger.warning(f"Parsed packet length "
+                               f"({actual_length_parsed}B) did not match "
+                               f"length specified in header ({n_bytes_packet}B). "
+                               f"Updating the position to the correct position "
+                               "indicated by CCSDS header.")
+                if not parse_bad_pkts:
+                    logger.warning("Skipping (not yielding) bad packet because parse_bad_pkts is falsy.")
```
greglucas, line 676:
```suggestion
                    logger.warning(f"Skipping (not yielding) bad packet with apid {header['PKT_APID'].raw_value.}")
```

greglucas (commented):
I realize you just copy-pasted a lot of this over, so sorry for the loads of comments here. But it seems worthwhile to clean some of it up and hopefully won't be hard to address most of them.

Overall this looks great to me and I like this approach!

space_packet_parser/definitions.py
```diff
@@ -1,20 +1,49 @@
...
+
+    def __init__(self, *args, partial_data: dict = None):
+        """
+        Parameters
+        ----------
+        partial_data : dict, Optional
+            Data parsed so far (for debugging at higher levels)
+        """
+        super().__init__(*args)
+        self.partial_data = partial_data
```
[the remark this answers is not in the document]
medley56: I think it's valuable to get back a Packet object (even partially filled out) as part of the exception. It allows users to check the type of the yielded object (`isinstance(next_packet, Exception)`) and examine the packet data parsed up to that point, which is extremely useful in debugging possibly faulty packet definitions. If we just attach it to the message, it's harder to examine the partial packet contents.u

Something I do want to add though is a second exception type for length mismatch that behaves much the same way.

examples/parsing_and_plotting_idex_waveforms_from_socket.py
```diff
@@ -110,7 +110,7 @@ def plot_full_event(data: dict):
     p.start()
 
     # Create a packet generator that listens to a socket
-    idex_packet_generator = idex_parser.generator(receiver)
+    idex_packet_generator = idex_parser.packet_generator(receiver)
```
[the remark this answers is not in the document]
medley56: Isn't it annoying that you can only make suggested changes to lines that have already changed? It drives me nuts.

space_packet_parser/definitions.py
```diff
@@ -43,13 +77,19 @@ def __init__(self, xtce_document: TextIO, ns: Optional[dict] = None):
         xtce_document : TextIO
             Path to XTCE XML document containing packet definition.
         ns : Optional[dict]
-            Optional different namespace than the default xtce namespace.
+            Optional different namespace than the namespace defined in the XTCE document.
+        word_size : Optional[int]
```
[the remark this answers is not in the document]
medley56: Removing `word_size`

space_packet_parser/definitions.py
medley56: Good catch!

space_packet_parser/definitions.py
medley56: I think this is probably possible and would be another nice performance enhancement but it has enough impact that I don't really want to tackle that on this PR. 

I think it would basically require us to be able to split the calls to `parse_CCSDSPacket` into two separate calls -- one to parse the header and a second to continue the parsing for the rest of the packet. We would need to rely on field ordering in order to pull out the necessary packet length field. 

Also, you could envision a pathological case where a user has decided they don't really care about the header information so they just define the header as a single blob of 6 bytes rather than splitting it into distinct fields, which would be valid XTCE but would break the ability to grab the packet length field out of the user-defined header representation.

greglucas: Yes, sounds good to leave this for the future.

> Also, you could envision a pathological case where a user has decided they don't really care about the header information so they just define the header as a single blob of 6 bytes rather than splitting it into distinct fields, which would be valid XTCE but would break the ability to grab the packet length field out of the user-defined header representation.

I think we would fail there because we are reading in PKT_LEN bytes to create the "raw bytes to be parsed" object to pass to our XTCE definition parsing (i.e. one full packet's worth of bytes as defined by the header and no more/less). So there has to be at least a valid APID and PKT_LEN attribute in the header I think.

greglucas (approved):

docs/source/developers.md
```diff
@@ -17,6 +17,12 @@ To run all tests in docker containers (tests against many versions of python), r
 docker-compose up --build && docker-compose down
 ```
 
+## Building Documentation with Sphinx
+Documentation is automatically built on ReadTheDocs but you can also build it locally with:
+```bash
+make html
```
medley56, line 23:
```suggestion
# From docs directory
make html && open build/html/index.html
```

space_packet_parser/definitions.py
medley56: 💯 But if we require header items to be present as predictable, separate items, we are levying requirements on the XTCE that aren't in the spec. I think it would be completely valid for a user parsing a simple packet to just lump all the header items into a 6 byte binary blob and ignore it. I don't think that's common but I don't think it's any kind of violation to say "I don't care about the headers. My packets are fixed length and I only want the user data."

Maybe it's OK to levy those additional requirements for the library. What do you think?

space_packet_parser/definitions.py
```diff
@@ -385,269 +359,326 @@ def _get_container_base_container(
...
-            pertaining to a single container type.
-        pkt_apid_header_name : str
-            The string used in the packet header describing the APID for the CCSDS packet.
+        packet: packets.CCSDSPacket
+            Binary representation of the packet used to get the coming bits and any
+            previously parsed data items to infer field lengths.
+        containers : dict
+            Dictionary of named containers, including their inheritance information.
+        root_container_name : str, Optional
+            Default is CCSDSPacket. Any root container may be specified.
```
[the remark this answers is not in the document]
medley56: Yep. Done.

space_packet_parser/definitions.py
```diff
@@ -385,269 +359,326 @@ def _get_container_base_container(
...
+                buffer_read_size_bytes = 4096
+
+        # ========
+        # Packet loop. Each iteration of this loop yields a ParsedPacket object
+        # ========
+        start_time = time.time_ns()
+        n_bytes_parsed = 0  # Keep track of how many bytes we have parsed
+        n_packets_parsed = 0  # Keep track of how many packets we have parsed
+        read_buffer = b""  # Empty bytes object to start
+        skip_header_bytes = skip_header_bits // 8  # Internally keep track of bytes
```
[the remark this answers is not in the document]
medley56: Done.

PR #82, opened 2024-09-26
Parser read method

### LINKED ISSUE #80: Move isinstance checks up a level when reading raw bytes
Just writing this down for my memory later, but no need to do anything with this now.

Here I think we could put these `isinstance` checks outside this function so it is only checked once on the initial input stream and not every read.

```suggestion
            if isinstance(source, io.BufferedIOBase):
                byte_reader = source.read
            elif isinstance(source, socket.socket):
                byte_reader = source.recv
        ...
        data = byte_reader(read_size_bytes)
```

_Originally posted by @greglucas in https://github.com/medley56/space_packet_parser/pull/79#discussion_r1776261754_

greglucas:
# Title of PR

We can avoid the `isinstance` checks in the while-loops by explicitly setting the type of read to do on the initial entry into the generator.

closes #80 

## Checklist

### DIFF
--- a/space_packet_parser/definitions.py
+++ b/space_packet_parser/definitions.py
@@ -532,36 +532,8 @@ class XtcePacketDefinition:
             If yield_unrecognized_packet_errors is True, it will yield an unraised exception object,
             which can be raised or used for debugging purposes.
         """
-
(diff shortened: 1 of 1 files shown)


medley56 (approved):
Heck yeah! This is a nice improvement!

PR #86, opened 2024-10-02
MNT: Use built-in types for return data objects

greglucas:
This adds the ability to subclass the built-in types and return an item that can be acted upon directly rather than having to access `parsed_item.raw_value` as an attribute to do any work with the item.

The built-in types are immutable and can't have attributes added to them so we inject the `raw_value` attribute in the new constructor of each explicit type.

**this PR**

```
{'VERSION': 0, 'TYPE': 0, 'SEC_HDR_FLG': 1, 'PKT_APID': 1424, 'SEQ_FLGS': 3, 'SRC_SEQ_CTR': 77, 'PKT_LEN': 1065, 'SHCOARSE': 1343, 'SHFINE': 19201, ...
```

**main**
```
{'VERSION': ParsedDataItem(name='VERSION', raw_value=0, unit=None, derived_value=0, short_description=None, long_description=None), 'TYPE': ParsedDataItem(name='TYPE', raw_value=0, unit=None, derived_value=0, short_description=None, long_description=None), 'SEC_HDR_FLG': ParsedDataItem(name='SEC_HDR_FLG', raw_value=1, unit=None, derived_value=1, short_description=None, long_description=None), 'PKT_APID': ParsedDataItem(name='PKT_APID', raw_value=1424, unit=None, derived_value=1424, short_description=None, long_description=None), 'SEQ_FLGS': ParsedDataItem(name='SEQ_FLGS', raw_value=3, unit=None, derived_value=3, short_description=None, long_description=None), 'SRC_SEQ_CTR': ParsedDataItem(name='SRC_SEQ_CTR', raw_value=77, unit=None, derived_value=77, short_description=None, long_description=None), 'PKT_LEN': ParsedDataItem(name='PKT_LEN', raw_value=1065, unit=None, derived_value=1065, short_description=None, long_description=None), 'SHCOARSE': ParsedDataItem(name='SHCOARSE', raw_value=1343, unit='dn', derived_value=1343, short_description='Secondary Header Coarse Time', long_description='CCSDS Packet 2nd Header Coarse Time'), 'SHFINE': ParsedDataItem(name='SHFINE', raw_value=19201, unit='dn', derived_value=19201, short_description='Secondary Header Fine Time', long_description='CCSDS Packet 2nd Header Fine Time'), ...
```

## Checklist

### DIFF
--- a/space_packet_parser/definitions.py
+++ b/space_packet_parser/definitions.py
@@ -370,10 +370,8 @@ class XtcePacketDefinition:
         header = {}
         current_bit = 0
         for item in CCSDS_HEADER_DEFINITION:
-            header[item.name] = packets.ParsedDataItem(
-                name=item.name,
-                # pylint: disable=protected-access
-                raw_value=packets._extract_bits(packet_data, current_bit, item.nbits))
+            # pylint: disable=protected-access
+            header[item.name] = packets._extract_bits(packet_data, current_bit, item.nbits)
             current_bit += item.nbits
         return header
 
@@ -417,7 +415,7 @@ class XtcePacketDefinition:
                     raise UnrecognizedPacketTypeError(
                         f"Detected an abstract container with no valid inheritors by restriction criteria. This might "
                         f"mean this packet type is not accounted for in the provided packet definition. "
-                        f"APID={packet['PKT_APID'].raw_value}.",
+                        f"APID={packet['PKT_APID']}.",
                         partial_data=packet)
                 break
 
@@ -593,7 +591,7 @@ class XtcePacketDefinition:
             # per the CCSDS spec
             # 4.1.3.5.3 The length count C shall be expressed as:
             #   C = (Total Number of Octets in the Packet Data Field) – 1
-            n_bytes_data = header['PKT_LEN'].raw_value + 1
+            n_bytes_data = header['PKT_LEN'] + 1
             n_bytes_packet = CCSDS_HEADER_LENGTH_BYTES + n_bytes_data
 
             # Based on PKT_LEN fill buffer enough to read a full packet
@@ -623,17 +621,17 @@ class XtcePacketDefinition:
                 packet = self.parse_ccsds_packet(packet,
                                                  root_container_name=root_container_name)
             except UnrecognizedPacketTypeError as e:
-                logger.debug(f"Unrecognized error on packet with APID {header['PKT_APID'].raw_value}'")
+                logger.debug(f"Unrecognized error on packet with APID {header['PKT_APID']}'")
                 if yield_unrecognized_packet_errors:
                     # Yield the caught exception without raising it (raising ends generator)
                     yield e
                 # Continue to next packet
                 continue
 
-            if packet.header['PKT_LEN'].raw_value != header['PKT_LEN'].raw_value:
+            if packet.header['PKT_LEN'] != header['PKT_LEN']:
                 raise ValueError(f"Hardcoded header parsing found a different packet length "
-                                 f"{header['PKT_LEN'].raw_value} than the definition-based parsing found "
-                                 f"{packet.header['PKT_LEN'].raw_value}. This might be because the CCSDS header is "
+                                 f"{header['PKT_LEN']} than the definition-based parsing found "
+                                 f"{packet.header['PKT_LEN']}. This might be because the CCSDS header is "
                                  f"incorrectly represented in your packet definition document.")
 
             actual_length_parsed = packet.raw_data.pos // 8
@@ -644,7 +642,7 @@ class XtcePacketDefinition:
                                f"Updating the position to the correct position "
                                "indicated by CCSDS header.")
                 if not parse_bad_pkts:
-                    logger.warning(f"Skipping (not yielding) bad packet with apid {header['PKT_APID'].raw_value}.")
+                    logger.warning(f"Skipping (not yielding) bad packet with apid {header['PKT_APID']}.")
                     continue
 
             yield packet
--- a/space_packet_parser/encodings.py
+++ b/space_packet_parser/encodings.py
@@ -3,8 +3,8 @@
 from abc import ABCMeta, abstractmethod
 import logging
 import struct
-from typing import Any, List, Optional, Tuple, Union
-# Installed
+from typing import List, Optional, Union
+
 import lxml.etree as ElementTree
 # Local
 from space_packet_parser import calibrators, comparisons, packets
@@ -145,7 +145,7 @@ class DataEncoding(comparisons.AttrComparable, metaclass=ABCMeta):
         """
         raise NotImplementedError()
 
-    def parse_value(self, packet: packets.CCSDSPacket, **kwargs) -> Tuple[Any, Any]:
+    def parse_value(self, packet: packets.CCSDSPacket, **kwargs) -> packets.ParameterDataTypes:
         """Parse a value from packet data, possibly using previously parsed data items to inform parsing.
 
         Parameters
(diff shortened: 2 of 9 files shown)


space_packet_parser/encodings.py
```diff
@@ -486,24 +486,24 @@ def parse_value(self,
         Returns
         -------
         : any
```
medley56, line 488: Update the return type hints in the docstrings for the `parse` methods.

space_packet_parser/packets.py
```diff
@@ -2,34 +2,46 @@
...
 
 from dataclasses import dataclass, field
-from typing import Union, Optional, Protocol, List
+from typing import Any, Optional, Protocol, List
 
 
-@dataclass
-class ParsedDataItem:
-    """Representation of a parsed parameter
+class DataItem:
```
medley56, line 8: Gah I really want to call this `Parameter`, but alas...
Not for this PR but wondering if we should prefix all the XTCE UML model classes with `Xtce`.

space_packet_parser/packets.py
```diff
@@ -2,34 +2,46 @@
...
-        Raw representation of the parsed value. May be lots of different types but most often an integer
-    derived_value : float or str
-        May be a calibrated value or an enum lookup
-    short_description : str
-        Parameter short description
-    long_description : str
-        Parameter long description
+    We need to override the __new__ method to store the raw value of the data item
+    on immutable built-in types. So this is just a way of allowing us to inject our
+    own attribute into the built-in types.
```
medley56, line 13: This is where I was concerned about complexity being introduced. But looking at the code, it's really not complex. Glad Python has moved away from the advice not to subclass primitives.

space_packet_parser/packets.py
```diff
@@ -2,34 +2,46 @@
...
     """
-    name: str
-    raw_value: Union[bytes, float, int, str]
-    unit: Optional[str] = None
-    derived_value: Optional[Union[float, str]] = None
-    short_description: Optional[str] = None
-    long_description: Optional[str] = None
+    def __new__(cls, value: Any, raw_value: Any = None) -> Any:
```
medley56, line 15: Is this the signature we want on this class? I believe that every DataItem should have a raw_value, no matter what. The raw value should be the closest representation of the encoded data we can return.
e.g. `bytes` for strings and binary blobs. `int` for integers, etc. I think I want to write some brief documentation on the philosophy of these representations since there is no objective meaning of "raw" value. Again, not for this PR. I'll make an Issue.

space_packet_parser/parameters.py
```diff
@@ -257,14 +257,14 @@ def parse_value(self, packet: packets.CCSDSPacket, **kwargs):
         derived_value : str
```
medley56, line 257: Fix docstring type hinting and add type hint to function signature. This goes for all the `parse_value` methods.

space_packet_parser/definitions.py
```diff
@@ -370,10 +370,8 @@ def _parse_header(packet_data: bytes) -> dict:
         header = {}
         current_bit = 0
         for item in CCSDS_HEADER_DEFINITION:
-            header[item.name] = packets.ParsedDataItem(
-                name=item.name,
-                # pylint: disable=protected-access
-                raw_value=packets._extract_bits(packet_data, current_bit, item.nbits))
+            # pylint: disable=protected-access
+            header[item.name] = packets._extract_bits(packet_data, current_bit, item.nbits)
```
medley56, line 374: Remind me why `_extract_bits` is a private function?

medley56 (changes requested):
Typehinting needs some cleanup in both docstrings an signatures. Looks good otherwise I think. I had a few questions in there and I'm going to make a couple of follow-on small issues.

space_packet_parser/packets.py
medley56: If you agree with #88 , you can incorporate it here, or not.

greglucas: I made this a private `_Parameter` as it really shouldn't be used by other people, it is for internal use.

I updated everything to `StrParameter` for the naming now. Is that what you had in mind? I did not modify any of the other "Parameter" entries.

It is a value here though and not a "parameter" right? So do we want to add or include "Value" or "Data" along with Parameter?

space_packet_parser/packets.py
greglucas: I added your text to the class docstring. I also made this a private class now to only be used internally.

space_packet_parser/definitions.py
greglucas: Do we want users to rely on this function? We can make it public if so, but I figured this was internal implementation details.

space_packet_parser/packets.py
medley56: I think semantically most people will assume a Parameter means a value but I'd also be happy with StrParameterValue if you think that's clearer.

space_packet_parser/definitions.py
medley56: I guess it's a difference of convention between private to the module vs private to the library. My understanding is that a _private_method is private to the module, which is why pylint is complaining that we're accessing a private function outside of the module in which it's defined.

I don't think users need it but the convention says we shouldn't be using it, even in a different module. But maybe I misunderstand that convention.

greglucas: I have the different take on it from pylint. The `_private()` is just to indicate it is something that users should not rely on and can be changed or removed at any point without a deprecation. This is the only place we use it outside of the module it is defined in, so maybe there is a better way to handle this, or maybe that is just to make this helper function public?

space_packet_parser/parameters.py
```diff
@@ -312,15 +310,13 @@ def parse_value(self, packet: packets.CCSDSPacket, **kwargs):
...
-        derived_value : str
+        derived_value : BoolParameter
             Resulting boolean representation of the encoded raw value
         """
-        raw, _ = super().parse_value(packet, **kwargs)
+        raw = super().parse_value(packet, **kwargs)
         # Note: This behaves very strangely for String and Binary data encodings.
         # Don't use those for Boolean parameters. The behavior isn't specified well in XTCE.
-        return raw, bool(raw)
+        return packets.BoolParameter(bool(raw), raw)
```
greglucas, line 319: @medley56 I did have to update a test for this which I wanted to get your take on.

The `raw` here is actually the "derived value" since it is the transformed return. So we are doing `bool(derived_value)` now. Is that what we want, or do we want to have it as:
```python
raw = super().parse_value(packet, **kwargs).raw_value
```

It seems like you'd want the derived value to produce the boolean of? i.e. if I have a calibrated value that is `raw + 1` and my raw value is `0`. I'd want that to be True and not False.

1. `bool(value), value`: current implementation, tells a user what was converted to the boolean
2. `bool(value), value.raw_value`: gives the user the most "raw" representation, but avoids the intermediary encoding
3. `bool(value.raw_value), value.raw_value`: only ever uses the raw value and avoids any derived/calibrated values.

Which version should we use here?

space_packet_parser/definitions.py
medley56: I'm not stressed about it. LGTM

space_packet_parser/parameters.py
medley56: Oof, this is a good question. I think we want BooleanParameterType to take the bool of the derived value. The issue is that we have discovered a parameter type that has multiple intermediate representations from raw data --> raw value --> calibrated value --> bool. This is something I never noticed but now I wonder if we have the same problem for enumerated parameter types. 

Let's skip the calibrated (intermediate) value as you suggest in option 2.

greglucas: OK, I just pushed (2) up.

Re: Enumerated... We are only passing back StringParameters there now. Is there a way to have something else in XTCE like pointing to a different parameter lookup or to say "read this list as an int"? Something we can tackle in a follow-up issue and would be a bug-fix if anything.

medley56: #93 Will address this.

medley56 (approved):
Looks great!

Issue #93, opened 2024-10-03
Handle Boolean and Enumerated Intermediate Values


medley56:
For Boolean and Enumerated parameters, there is basically an intermediate form of the parsed value. The progression goes:

raw bytes --> raw parsed value (e.g. int) --> optional calibrated value --> boolean or enum lookup

This scheme is not accounted for in our current implementation. The resulting BoolParameter or StrParameter objects only have the `value` and the `raw_value`, so what becomes of the intermediate states.

Currently, for Boolean, we drop the possibly calibrated intermediate value and we return the raw parsed value and the final boolean value that is basically `bool(calibrate(parse(raw_bytes)))`.

Ideas for improvement:
1. Add another "stage" of data representation to our returned Parameter classes (the ones that subclass int, str, float, etc) to also include a calibrated value that sits between the final value and the raw parsed value. 
2. Mutually exclusive to 1. Simply document that we do not return the intermediate calibrated value and make sure that Enums and Bools handle this consistently. Also double check the XTCE docs to make sure that Enums are allowed to be based on calibrated values. They should be because the calibrator is defined on the encoding, not the parameter element. In particular check for use of `useCalibratedValue` in the Enum and Bool parameter elements.

greglucas:
I think we should return the value used to calculate the "bool" or "enumeration" as the `raw_value` attribute. That value could then be used to get the raw-raw value.

https://github.com/medley56/space_packet_parser/blob/6ce65618b33853d7ce1f00cb97c7fee73a740488/space_packet_parser/parameters.py#L321

If we return `value` here instead of `value.raw_value`, the user could still get back to the raw value themselves from the intermediate. Right now, there is no way to do that.

medley56:
Closing based on #94

PR #94, opened 2024-10-03
MNT: Change to returning intermediate values for enumerated and booean parameters

greglucas:
This gives the ability to get back to the raw value if desired still, but returns what was used for the "lookup" / bool conversion as that is likely what a user needs for the first debug step. Then they can get to the raw-raw value themselves if they really need it.

closees #93 

## Checklist


medley56 (approved):
This seems like a reasonable approach to me.

PR #106, opened 2024-10-09
Add documentation about raw vs normal values for parsed parameters

medley56:
Add support for float and string encoded enumerated lookups

## Checklist


greglucas (approved):
The code all looks good to me. A minor suggestion on docs.

As a side note, I would be amazed if someone in flight software wanted an enumerated float lookup, that seems very fragile... I can just imagine some floating point errors creeping into an exact lookup like that!

medley56:
I totally agree about the float encoded enum. It's an absolutely terrible idea because of the imprecise storage of float values but XTCE allows it so I want to support it but document that it's a terrible idea 😆

PR #107, opened 2024-10-10
MNT/ENH: Add a raw CCSDS packet generator

### LINKED ISSUE #81: Avoid reading the header twice during packet parsing
I was actually wondering about this check and whether we could avoid reading the first 7 items in the container definition and rely on the CCSDS header being correct. That way we could start the Packet class right away when we are doing the header above and then continue on here. Currently we are parsing the header twice essentially.

Could we check this condition in the definition itself somehow so that when someone is creating the definition from the XTCE we would raise immediately before parsing even began?

_Originally posted by @greglucas in https://github.com/medley56/space_packet_parser/pull/79#discussion_r1776293877_

### LINKED ISSUE #104: Parse Headers and Return Raw Packet Data Only with Empty XtcePacketDefinition
# Context

In some use cases, people will want to parse CCSDS headers, returning the binary data directly and not worrying about the user data definitions. They will not have an XTCE document, nor do they need one for this use case.

# Implementation Plan

- Allow instantiation of XtcePacketDefinition without an XTCE document.
- When we parse a packet stream for headers only, we should get back the raw packet data for the packet along with the parsed header. This may already be the case.
- Add a test for parsing packets without a packet definition.

greglucas:
This refactors out the reading of a byte-stream into a separate generator function that can yield raw CCSDSPackets with only the raw bytes. The definitions and parsing logic can use this generator to yield packets and then parse whatever they need to after the fact. This enables someone to use this generator without a packet definition and to investigate what is in the packet stream without needing to know more detailed information about how individual packets are laid out internally.

closes #81
closes #104

## Checklist

### DIFF
--- a/space_packet_parser/packets.py
+++ b/space_packet_parser/packets.py
@@ -295,7 +295,7 @@ class CCSDSPacket(dict):
 
 class Parseable(Protocol):
     """Defines an object that can be parsed from packet data."""
-    def parse(self, packet: CCSDSPacket, **parse_value_kwargs) -> None:
+    def parse(self, packet: CCSDSPacket) -> None:
         """Parse this entry from the packet data and add the necessary items to the packet."""
 
 
@@ -336,13 +336,13 @@ class SequenceContainer(Parseable):
         self.restriction_criteria = self.restriction_criteria or []
         self.inheritors = self.inheritors or []
 
-    def parse(self, packet: CCSDSPacket, **parse_value_kwargs) -> None:
+    def parse(self, packet: CCSDSPacket) -> None:
         """Parse the entry list of parameters/containers in the order they are expected in the packet.
 
         This could be recursive if the entry list contains SequenceContainers.
         """
         for entry in self.entry_list:
-            entry.parse(packet=packet, **parse_value_kwargs)
+            entry.parse(packet=packet)
 
 
 def ccsds_generator(  # pylint: disable=too-many-branches,too-many-statements
--- a/space_packet_parser/packets.py
+++ b/space_packet_parser/packets.py
@@ -2,6 +2,8 @@
 """Packet containers and parsing utilities for space packets."""
 
 from dataclasses import dataclass, field
+from enum import IntEnum
+from functools import cached_property
 from typing import List, Optional, Protocol, Union
 
 BuiltinDataTypes = Union[bytes, float, int, str]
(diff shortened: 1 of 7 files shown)


medley56:
Nice! I like this. My current thinking is that I prefer this to your alternative idea because I want to avoid changing the user API unnecessarily and I think the use cases for examining packets as binary data only are mostly going to use an additional abstraction layer rather than users calling the `packet_generator` directly. I think we should leave it public in case someone decides they want to use it in a custom way but I think `packet_generator` is mostly going to be used through CLI tools.

greglucas:
I think we could support backwards compatibility with the `definition.packet_generator()` if we want and use that to redirect to the primary packet generator location with a deprecation if we want too.

```python
        definition = None if ccsds_headers_only else self
        for packet in packets.packet_generator(binary_data,
                                               definition=definition,
                                               buffer_read_size_bytes=buffer_read_size_bytes,
                                               show_progress=show_progress,
                                               skip_header_bytes=skip_header_bytes):
            yield packet
```

The benefit I see is that there is only one generator that someone would have to know about and they could bring their own definition(s) to it. I'm envisioning that we have multiple XTCE definitions hanging around on a mission and don't want to potentially have to know which definition I need to use, so create a list and iterate through it. (Is there a way to combine XTCE definitions internally for us to consider as well?)

```python
definitions = [XtcePacketDefinition(file) for file in list_of_files]
```

Currently I think we'd have to do this:
```python
for definition in definitions:
    generator = getattr(definition, "packet_generator")
    for packet in generator(data):
        # Do something
```

With a single entry-point we could write it like this:
```python
for definition in definitions:
    for packet in packets.packet_generator(data, definition=definition):
        # Do something
```

Thinking towards the future, we could even potentially support passing in multiple definitions with this approach and doing the iteration for a user:
```python
for packet in packets.packet_generator(data, definition=definitions):
    # Do something
```

greglucas:
I pushed two new commits. One somewhat unrelated but was easier to clean it up here by removing the `**parse_kwargs` that appear unused now. I can move that to a separate PR if desired. The final commit moves all of the packet parsing logic into the `packets.py` file now, but leaves the `definition.packet_generator()` around and just yields from the main generator passing through the proper arguments. Let me know your thoughts on those and I can break them off into separate PRs if you'd prefer to keep this just on adding the new functionality for now.

space_packet_parser/packets.py
```diff
@@ -56,28 +73,87 @@ class StrParameter(_Parameter, str):
...
+                f"{self.secondary_header_flag=}, {self.apid=}, {self.sequence_flags=}, "
+                f"{self.sequence_count=}, {self.data_length=})").replace("self.", "")
+
+    @cached_property
+    def version_number(self) -> int:
+        """CCSDS Packet Version Number"""
+        return _extract_bits(self, 0, 3)
+
+    @cached_property
+    def type(self) -> int:
```
greglucas, line 107: `type` is a built-in function in Python, so do we care that this is called `RawPacketData.type`? Should we prefix it with something like `packet_type` instead? To me that is redundant, but happy to change if others want a different naming.

medley56: I think this is fine for a property method

medley56 (approved):

medley56 (approved):

PR #108, opened 2024-10-16
Move packet header contents onto the CCSDSPacket object

greglucas:
These constants are directly tied to the CCSDSPacket definition, so put these onto the class itself for use.

Note that this builds off of the `packet_generator` PR, so only the final commit is relevant to look at here for now.

## Checklist


greglucas:
Closing this because I've refactored the header into properties in the packet generator refactor which is better than this approach.

PR #111, opened 2024-10-21
ENH: Add an Excel spreadsheet to XTCE conversion utility

greglucas:
This adds a utility function that converts spreadsheet definition files (GSEOS format?) into XTCE definitions that space_packet_parser can read and use.

closes https://github.com/medley56/space_packet_parser/discussions/74

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,9 @@ optional = true
 matplotlib = ">=3.4"
 memory-profiler = "^0.61.0"
 
+[tool.poetry.scripts]
+spp-convert = "space_packet_parser.converters:main"
+
 [build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
--- a/tests/test_data/expected_excel_xtce.xml
+++ b/tests/test_data/expected_excel_xtce.xml
@@ -1,0 +1,171 @@
+<?xml version='1.0' encoding='utf-8'?>
+<xtce:SpaceSystem xmlns:xtce="http://www.omg.org/space/xtce" name="Test Instrument">
+	<xtce:Header date="2024-07-26 00:00:00" version="v1.2" author="Space Packet Parser" />
+	<xtce:TelemetryMetaData>
+		<xtce:ParameterTypeSet>
+			<xtce:IntegerParameterType name="VERSION" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="3" encoding="unsigned" />
+			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="TYPE" signed="false">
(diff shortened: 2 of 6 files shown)


tests/test_data/expected_excel_xtce.xml
```diff
@@ -0,0 +1,171 @@
...
+			<xtce:IntegerParameterType name="SEQ_FLGS" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="2" encoding="unsigned" />
+			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="SRC_SEQ_CTR" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="14" encoding="unsigned" />
+			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="PKT_LEN" signed="false">
+				<xtce:IntegerDataEncoding sizeInBits="16" encoding="unsigned" />
+			</xtce:IntegerParameterType>
+			<xtce:IntegerParameterType name="TEST_PACKET.SHCOARSE" signed="false">
```
greglucas, line 27: I'm not sure if you have a better way to handle this, but we have potentially duplicate names across packets with different lengths/datatypes. For example, someone may want to say packet1 has "VAR1 == uint3", but packet2 has "VAR1 == uint8" (especially relevant for spare/fill that are different beetween packets). So, how do we represent that in the XTCE? On IMAP we chose to prepend the `packetName.mnemonic` to avoid this, but it does add a decent amount of bloat and then does not allow someone to directly go in and say `packet["mnemonic"]`, rather they have to do `packet["packet_name.mnemonic"]`. We strip that off ourselves, which works OK for us. But it would be good to hear if there is something we are missing or could be doing better.

pyproject.toml
```diff
@@ -58,6 +58,9 @@ optional = true
 matplotlib = ">=3.4"
 memory-profiler = "^0.61.0"
 
+[tool.poetry.scripts]
+spp-convert = "space_packet_parser.converters:main"
```
greglucas, line 62: If we have a different command line utility named `spp`, do we want this to be a subcommand like `spp convert /path/to/file` instead of a totally separate CLI?

tests/test_data/expected_excel_xtce.xml
medley56: I think this approach makes sense. I would probably double-underscore it rather than dot it but that's only because I've seen FSW do that. In fact they tend to massively duplicate identical type definitions as a result when exporting XTCE. It bloats the file but doesn't break anything.

greglucas: I think we had something with double unders causing issues. We should make a separate flag or something to use, or even split on both of those to catch both cases...

medley56:
After our discussion with SWRI I think we are going another direction but this could be reworked as another example of a conversion to XTCE.

PR #113, opened 2024-11-10
PERF: Reduce number of times findall is called on XML file

### LINKED ISSUE #112: Reading the XTCE file slower than parsing
# Context

The initial XML reading to create objects can be slower than the actual packet parsing when there are a lot of parameters and items in the file. This is relevant for the CTIM file which has lots of parameters, a quick example script for testing:

```python
from pathlib import Path
import time

from space_packet_parser import definitions

xtce_document = Path("tests/test_data/ctim/ctim_xtce_v1.xml")

start = time.time()
packet_definition = definitions.XtcePacketDefinition(xtce_document)
total_time = time.time() - start
print(f"It took {total_time:.2f} seconds to parse the XTCE document")

print("Number of containers:", len(packet_definition._sequence_container_cache))
print("Number of parameters:", len(packet_definition._parameter_cache))
print("Number of parameter types:", len(packet_definition._parameter_type_cache))
```

on main, it currently takes ~12.8s
```
It took 12.79 seconds to parse the XTCE document
Number of containers: 39
Number of parameters: 9491
Number of parameter types: 15
```

Profiling this, the slowdown in this case is in the `_find_parameter()` method. In there, an XML `findall()` call is issued for every parameter.

# Implementation Plan

The `findall()` can be called just once to update the cache with all of the items rather than for each of the parameters iteratively as we go through the list. I propose to add a new `_populate_parameter_cache()` method that will get called during initialization, and then future lookups happen based on the cache rather than the `_find_parameter()` method.

greglucas:
The parameter types and parameters can be read through once and loaded into a cache straight away and then the lookups done on the cache after that. This prevents a findall() call iterating through the entire XML document every time a new parameter is found.

Profiling with the script provided in #112, we go from 12s to 0.03s for the Definition creation.

closes #112

## Checklist

### DIFF
--- a/space_packet_parser/definitions.py
+++ b/space_packet_parser/definitions.py
@@ -87,6 +87,8 @@ class XtcePacketDefinition:
         self.type_tag_to_object = {k.format(**self.ns): v for k, v in
                                    self._tag_to_type_template.items()}
 
+        self._populate_parameter_type_cache()
+        self._populate_parameter_cache()
         self._populate_sequence_container_cache()
 
     def __getitem__(self, item):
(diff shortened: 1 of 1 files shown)


greglucas:
Ping for a review here. I think this would really help IMAP as some of those XTCE files can be quite large and this should really speed up initial parsing without change in functionality.

medley56 (approved):

PR #118, opened 2025-01-30
Modify IMAP code to create generic xarray dataset from packet data file

medley56:
## Checklist

This is a first stab at a working implementation based on the IMAP code. I removed anything that was particular to IMAP, including the time dimension.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -25,6 +25,8 @@ Release notes for the `space_packet_parser` library
     ``definition.packet_generator(data, combine_segmented_packets=True, secondary_header_bytes=4)``
 - Add a command line interface (spp) to enable quick and easy access to
   some common tasks and utilities.
+- Add function to directly create an `xarray.DataSet` from a packet file and XTCE definition.
+  e.g. `space_packet_parser.xarr.create_dataset([packets1, packets2, ...], definition)`
 
 ### v5.0.1 (released)
 - BUGFIX: Allow raw_value representation for enums with falsy raw values. Previously these defaulted to the enum label.
--- a/tests/integration/test_xarr.py
+++ b/tests/integration/test_xarr.py
@@ -1,0 +1,53 @@
+"""Test creating an xarray dataset from CCSDS packets"""
+import pytest
+pytest.importorskip("xarray")
+pytest.importorskip("numpy")
+import numpy as np
+from space_packet_parser.xarr import create_dataset
+
+
+def test_create_xarray_dataset(jpss_test_data_dir):
+    """Test creating an xarray dataset from JPSS geolocation packets"""
+    packet_file = jpss_test_data_dir / "J01_G011_LZ_2021-04-09T00-00-00Z_V01.DAT1"
+    definition_file = jpss_test_data_dir / "jpss1_geolocation_xtce_v1.xml"
+    ds = create_dataset(packet_file, definition_file)
+    assert list(ds.keys()) == [11]
+    assert len(ds[11]) == 27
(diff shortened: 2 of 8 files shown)


tests/integration/test_xarr.py
greglucas, line 1: Add some dtype checks to verify the numpy conversions are as expected, some raw/derived value parameterizations too? Examples in IMAP's codebase:

https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/1281c93b1b976ef705d14e8a6f67064fa16be992/imap_processing/tests/test_utils.py#L78

tests/integration/test_xarr.py
```diff
@@ -0,0 +1,22 @@
+"""Test creating an xarray dataset from CCSDS packets"""
```
greglucas, line 1: `pytest.importorskip("xarray")` ?

space_packet_parser/extras/xarr.py
greglucas, line 1: Do we anticipate other extras, or would this make sense to bring up into the top-level as another module there?

space_packet_parser/extras/xarr.py
```diff
@@ -0,0 +1,196 @@
...
+    use_raw_value : bool
+        Default False. If True, uses the data type of the raw value for each parameter.
+
+    Returns
+    -------
+    datatype : Optional[str]
+        The minimum numpy dtype for the parameter.
+        Returns None to indicate that numpy should use default dtype inference.
+    """
+    data_encoding = definition.named_parameters[name].parameter_type.encoding
```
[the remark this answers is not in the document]
medley56: If we use ParameterType, we are using the "derived" value type. e.g. for an integer encoded Enum parameter, that's ultimately a string. But in that case, we don't know what the raw value type (encoding type) was, likely a uint.

If we use the encoding from the XTCE, we have the reverse problem. We know the encoded type but we don't know what type the final value is (hence all the `return None`). 

Maybe the solution is both? i.e. if the user is asking for raw values, we use the encoding from the definition. If they want the final values, we look at the type of the value coming out of the parser.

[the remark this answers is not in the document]
greglucas: Oh that is a good point. I think I agree with you that handling both makes sense here depending on what is requested.

space_packet_parser/extras/xarr.py
medley56: I'll move it up. If there are other extras later, we can always refactor.

space_packet_parser/extras/xarr.py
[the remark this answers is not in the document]
medley56: I think this is handled. There's only so much we can do for derived values without doing a lot of picking apart the XTCE in great detail.

tests/integration/test_xarr.py
medley56: Done.

docs/source/changelog.md
```diff
@@ -25,6 +25,7 @@ Release notes for the `space_packet_parser` library
     ``definition.packet_generator(data, combine_segmented_packets=True, secondary_header_bytes=4)``
 - Add a command line interface (spp) to enable quick and easy access to
   some common tasks and utilities.
+- Add function to directly create an `xarray.DataSet` from a packet file and XTCE definition.
```
greglucas, line 28: Call it out directly with a quick example?
```suggestion
- Add function to directly create an `xarray.DataSet` from a packet file and XTCE definition, `space_packet_parser.extras.xarr.create_dataset()`
```

greglucas (approved):

PR #124, opened 2025-02-04
MNT: Switch linting to ruff

greglucas:
This switches over from pycodestyle + pylint to ruff. The codes aren't exactly 1-1, but we can add more/less as desired (there are lots of options https://docs.astral.sh/ruff/rules/). I tried to just use some basic ones for now that keeps it pretty similar. One nice thing is the addition of pyupgrade checks that caught some unneeded typing usages where we can use the builtins now.

On IMAP, we use pre-commit.ci for our linting checks instead of a GitHub Action to separately tell us whether it is linting or tests failing. We can add that and add some more rules like codespell later if those are desired too.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/.pre-commit-config.yaml

You can run `ruff check --fix space_packet_parser` to automatically fix many issues.

Some notes:
* Imports are sorted
* Removed assertions within main codebase and changed to raising errors now
* Changed pytest to automatically error on warnings

## Checklist


medley56 (approved):
Looks good!

PR #125, opened 2025-02-04
CI: Add code coverage reports to external source

greglucas:
I think you need to enable codecov (or some other external service if you prefer) for this to work as I couldn't find a way to enable this. Maybe this is impetus to move to the LASP org so we can both admin things?
https://app.codecov.io/gh

On IMAP we use a token, but I don't think that is required anymore and it would be nice if we can get OIDC to work instead.
https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/693fbdfffd12fa3ec9249f2d5f03fbee39ae5039/.github/workflows/test.yml#L54-L57

I can add a badge once the coverage link/report is enabled.

## Checklist

### DIFF
--- a/.github/workflows/pr_tests.yml
+++ b/.github/workflows/pr_tests.yml
@@ -36,8 +36,12 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install ".[xarray]" pytest pytest-randomly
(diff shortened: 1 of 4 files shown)


medley56 (approved):
Nice! I'm looking forward to getting more familiar with the coverage app on Github.

PR #127, opened 2025-02-05
BUGFIX: update allowed float encodings to match XTCE standard

cgobat:
## Checklist
  - N/A

This PR resolves the issue I raised in my comment [here](https://github.com/medley56/space_packet_parser/pull/68#discussion_r1942092917).

### DIFF
--- a/space_packet_parser/encodings.py
+++ b/space_packet_parser/encodings.py
@@ -1,7 +1,7 @@
 """DataEncoding definitions"""
 # Standard
-import struct
 import logging
+import struct
 import warnings
 from abc import ABCMeta, abstractmethod
 from typing import Optional, Union
--- a/space_packet_parser/encodings.py
+++ b/space_packet_parser/encodings.py
@@ -603,14 +603,14 @@ class IntegerDataEncoding(NumericDataEncoding):
 
 class FloatDataEncoding(NumericDataEncoding):
     """<xtce:FloatDataEncoding>"""
-    _supported_encodings = ['IEEE-754', 'MIL-1750A']
+    _supported_encodings = ['IEEE754_1985', 'IEEE754', 'MILSTD_1750A']
     _data_return_class = packets.FloatParameter
 
     def __init__(
             self,
             size_in_bits: int,
             *,
-            encoding: str = 'IEEE-754',
+            encoding: str = 'IEEE754',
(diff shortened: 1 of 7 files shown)


space_packet_parser/encodings.py
```diff
@@ -632,18 +633,21 @@ def __init__(
             List of ContextCalibrator objects, containing match criteria and corresponding calibrators to use in
             various scenarios, based on other parameters.
         """
-        if encoding not in self._supported_encodings:
+        if encoding not in self._allowed_encodings:
             raise ValueError(f"Invalid encoding type {encoding} for float data. "
-                             f"Must be one of {self._supported_encodings}.")
-        if encoding == 'MIL-1750A' and size_in_bits != 32:
+                             f"Must be one of {self._allowed_encodings}.")
+        if encoding not in self._supported_encodings:
```
greglucas, line 639: Do we want to provide a softer landing for previous incorrect versions and nudge them towards the correct names?

I think we could add something like:
```python
if encoding == "IEEE-754":
    warnings.warn("IEEE-754 (with a dash) is not supported by the XTCE spec, use 'IEEE754' instead")
    encoding = "IEEE754"
elif same_for_mil:
    ...
else:
    raise NotImplemented

Comment (cgobat) [reply space_packet_parser/encodings.py:639]: Oh yeah, that's a great idea.

**Comment by cgobat** [comment]
Could you run tests/CI on this again?

**Comment by medley56** [comment]
@cgobat I just added you as a Collaborator. You should be able to run the CI tests now.

File: space_packet_parser/encodings.py [hunk 2]
```diff
@@ -1,7 +1,8 @@
 """DataEncoding definitions"""
 # Standard
-import logging
 import struct
+import logging
+import warnings
```
greglucas, line 5:
```suggestion
import logging
import struct
import warnings
```

greglucas (approved):
Thanks for adding in that soft deprecation of the wrong labels. I think this looks great!

There is one linting fix which you can get locally with `ruff --fix space_packet_parser` I think.

greglucas:
@cgobat, thanks! Also as an FYI, we just moved this over to the "lasp" GitHub organization, so you may want to update your upstream location when pulling/rebasing.

Issue #129, opened 2025-02-05
Add a schema validator / checker


greglucas:
It would be helpful to add a schema validation step or point to advice for how to do this for new users just setting up their initial XTCE documents. There is an official xsd document: https://www.omg.org/spec/XTCE/20180204/SpaceSystem.xsd
This would also catch incorrect implementations on our end, like with the float naming identified in #128 

A quick search shows there are some packages that do this, although I have no idea how robust/useful they would be: https://xmlschema.readthedocs.io/en/latest/usage.html

Some questions to think about:
* Do we want to add a dependency for schema validation, or just point to an external tool to do this. Could be a package extra too and not on by default.
* Do we want to commit the official xsd to the repository for checking? I have had that site go down on me several times now...

cgobat:
I agree that this would be a useful addition. I know that [lxml has some XSD-based validation capabilities](https://lxml.de/validation.html#xmlschema) already. Since lxml is already a dependency, maybe we can look at whether it would suffice?

greglucas:
That looks great if we can leverage what we already have as a dependency!

cgobat:
Does it make the most sense to automatically try to validate against the schema upon instantiation of the `XtcePacketDefinition` class?

greglucas:
> Does it make the most sense to automatically try to validate against the schema upon instantiation of the XtcePacketDefinition class?

Probably if it is a quick check, we already have the built-in dependency, and it doesn't raise on valid but complex schemas. It would be good to make sure wherever we add it that we add useful information saying _why_ the schema match failed.

medley56:
Here is a little example of doing schema validation with lxml. TIL that `author` is not a valid field in the XTCE Header and that `validationStatus` is a required field...

```
from lxml import etree
import requests

defdoc = "tests/test_data/jpss/jpss1_geolocation_xtce_v1.xml"

xsi_uri = "http://www.w3.org/2001/XMLSchema-instance"

tree = etree.parse(defdoc)

root = tree.getroot()
print(root.attrib)

# This attribute must always be a pair of strings separated by a space: e.g.
# xsi:schemaLocation="http://www.omg.org/spec/XTCE/20180204 https://www.omg.org/spec/XTCE/20180204/SpaceSystem.xsd"
xsd, xsd_url = root.attrib[f"{{{xsi_uri}}}schemaLocation"].split()

xsd_doc = requests.get(xsd_url)
#print(xsd_doc.content)

[...]

PR #136, opened 2025-02-12
Ensure uniqueness of parsed SequenceContainer objects

medley56:
Modify the parsing of XTCE documents to ensure that SequenceContainer objects are kept unique and reused instead of parsed and stored multiple times.

This PR adds checks to the parsing of ParameterSet, ParemeterTypeSet, and ContainerSet to ensure that the objects we parse out of those elements are uniquely represented in the resulting definition object. We also issue warnings for parameters and parameter types that are defined but not used.

## Checklist

### DIFF
--- a/space_packet_parser/common.py
+++ b/space_packet_parser/common.py
@@ -42,6 +42,7 @@ class XmlObject(metaclass=ABCMeta):
             tree: Optional[ElementTree.ElementTree],
             parameter_lookup: Optional[dict[str, any]],
             parameter_type_lookup: Optional[dict[str, any]],
+            container_lookup: Optional[dict[str, any]],
     ) -> 'XmlObject':
         """Create an object from an XML element
 
@@ -63,6 +64,8 @@ class XmlObject(metaclass=ABCMeta):
             Parameters dict for parsing that requires knowledge of existing parameters
         parameter_type_lookup: Optional[dict[str, parameters.ParameterType]]
             Parameter type dict for parsing that requires knowledge of existing parameter types
+        container_lookup: Optional[dict[str, parameters.ContainerType]]
+            Container type dict for parsing that requires knowledge of existing containers
 
         Returns
         -------
--- a/space_packet_parser/xtce/containers.py
+++ b/space_packet_parser/xtce/containers.py
@@ -1,6 +1,6 @@
 """Module with XTCE models related to SequenceContainers"""
 from dataclasses import dataclass, field
-from typing import Optional
+from typing import Optional, Union
 
 from lxml import etree as ElementTree
 from lxml.builder import ElementMaker
@@ -29,18 +29,18 @@ class SequenceContainer(common.Parseable, common.XmlObject):
         be included.
     abstract : bool
         True if container has abstract=true attribute. False otherwise.
-    inheritors : list, Optional
+    inheritors : Optional[list]
         List of SequenceContainer objects that may inherit this one's entry list if their restriction criteria
         are met. Any SequenceContainers with this container as base_container_name should be listed here.
     """
     name: str
-    entry_list: list  # List of Parameter objects, found by reference
+    entry_list: list[Union[parameters.Parameter, 'SequenceContainer']]
     short_description: Optional[str] = None
     long_description: Optional[str] = None
     base_container_name: Optional[str] = None
-    restriction_criteria: Optional[list] = field(default_factory=lambda: [])
+    restriction_criteria: Optional[list[comparisons.MatchCriteria]] = field(default_factory=lambda: [])
     abstract: bool = False
-    inheritors: Optional[list['SequenceContainer']] = field(default_factory=lambda: [])
+    inheritors: Optional[list[str]] = field(default_factory=lambda: [])
 
     def __post_init__(self):
         # Handle the explicit None passing for default values
@@ -63,6 +63,7 @@ class SequenceContainer(common.Parseable, common.XmlObject):
             ns: dict,
             tree: ElementTree.ElementTree,
             parameter_lookup: dict[str, parameters.Parameter],
+            container_lookup: Optional[dict[str, any]],
             parameter_type_lookup: Optional[dict[str, parameter_types.ParameterType]] = None
     ) -> 'SequenceContainer':
         """Parses the list of parameters in a SequenceContainer element, recursively parsing nested SequenceContainers
@@ -79,7 +80,9 @@ class SequenceContainer(common.Parseable, common.XmlObject):
         ns : dict
             XML namespace dict
         parameter_lookup : dict[str, parameters.Parameter]
-            Parameters contained in the entrylists of sequence containers
+            Parameters contained in the entry lists of sequence containers
+        container_lookup: Optional[dict[str, SequenceContainer]]
+            Containers already parsed, used to sort out duplicate references
         parameter_type_lookup : Optional[dict[str, parameter_types.ParameterType]]
             Ignored.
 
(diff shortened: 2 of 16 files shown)


space_packet_parser/xtce/containers.py
```diff
@@ -111,15 +120,24 @@ def from_xml(
...
+                        name=entry.attrib['containerRef'],
+                        ns=ns
                     )
+                    nested_container = cls.from_xml(
+                            nested_container_element,
+                            tree=tree,
+                            parameter_lookup=parameter_lookup,
+                            container_lookup=container_lookup,
+                            ns=ns
+                        )
```
greglucas, line 138: Is there a way to change this logic at all and put the if/else into the `_find_container_by_name`? Make that a `_get_container(name, ...)` lookup that would check the cache first and it would also create the object if it isn't found. I'm not sure if that is feasible given it is a classmethod currently and doesn't have the `self.containers` information to lookup in.
There is one other place this is used, and I wans't sure it would be helpful or not to change over to `_get_container_by_name()` so just putting this out there for your thoughts.

space_packet_parser/xtce/definitions.py
```diff
@@ -74,16 +74,18 @@ def __init__(
         author : Optional[str]
             Optional author name to include in XML when serializing.
         """
-        self._parameter_type_cache = {}
-        self._parameter_cache = {}
-        self._sequence_container_cache = {}
+        self.parameter_types = {}
+        self.parameters = {}
+        self.containers = {}
```
greglucas, line 79: Is `containers` the right name for this, or do we want to follow the XTCE naming and keep it as `sequence_containers`? I'm indifferent, these are honestly just things used internally and will rarely be used by an end-user.

space_packet_parser/xtce/definitions.py
```diff
@@ -74,16 +74,18 @@ def __init__(
...
         if sequence_container_list:
             for sc in sequence_container_list:
-                self._parameter_type_cache.update({p.parameter_type.name: p.parameter_type for p in sc.entry_list})
-                self._parameter_cache.update({p.name: p for p in sc.entry_list})
-            self._sequence_container_cache.update({sc.name: sc for sc in sequence_container_list})
+                # TODO: entry_list can contain SequenceContainers too (parseables) so this needs an
+                #  isinstance check.
+                self.parameter_types.update({p.parameter_type.name: p.parameter_type for p in sc.entry_list})
+                self.parameters.update({p.name: p for p in sc.entry_list})
+            self.containers.update({sc.name: sc for sc in sequence_container_list})
```
greglucas, line 88: You are looping multiple times over the same lists, so maybe just do things explicitly with nested loops for simplicity?
```suggestion
                # entry_list can contain other containers, so filter those out and only add parameters
                for p in sc.entry_list if isinstance(p, parameters.Parameter):
                    if p.name not in self.parameters:
                        self.parameters[p.name] = p
                    if p.parameter_type.name not in self.parameter_types:
                        self.parameter_types[p.parameter_type.name] = p.parameter_type
                if sc.name not in self.containers:
                    self.containers[sc.name] = sc
```

space_packet_parser/xtce/parameter_types.py
```diff
@@ -238,7 +241,8 @@ def from_xml(
             ns: dict,
             tree: Optional[ElementTree.Element] = None,
             parameter_lookup: Optional[dict[str, any]] = None,
-            parameter_type_lookup: Optional[dict[str, any]] = None
+            parameter_type_lookup: Optional[dict[str, any]] = None,
+            container_lookup: Optional[dict[str, any]] = None
```
greglucas, line 245: These seem repetitive if they are rarely used. Would it make sense to pass in a `definition` here instead so you can consolidate all 3 of those items onto the parent object that holds all of those. That may make testing harder though?

space_packet_parser/xtce/definitions.py
medley56: I chose `containers` over `sequence_containers` because it's more concise and it contains the objects parsed out of the `<ContainerSet>` element. I just checked the element description and `SequenceContainer` is the only valid child of `ContainerSet` so I think we could use either semantic.

space_packet_parser/xtce/definitions.py
medley56: I actually already reworked this basically as you suggest, in order to raise exceptions for `Parameter`, `ParameterType`, and `SequenceContainer` objects of the same name that are equal (`a == b`) but are not the same object in memory (`a is b`).

space_packet_parser/xtce/parameter_types.py
medley56: Yeah, these are all on the ABC to satisfy Liskov substitution. The argument for keeping these kwargs separate is to make it explicit which methods use which of the three lookups. And yes, it does make things easier to test if we can just pass lists rather than a full XtcePacketDefinition object.

space_packet_parser/xtce/containers.py
medley56: I'm kind of trying to keep the finding of an Element separate from the parsing of that element. I've renamed these methods to `_get_base_container_element` and `_get_container_element` to make it clear they are just finding XML Element objects in the tree. `from_xml` is responsible for deciding if we actually want to parse those elements and store them in `container_lookup`.

There are probably more optimizations to make here but I want to move on for now.

tests/integration/test_xtce_based_parsing/test_inheritance_restrictions.py
```diff
@@ -9,6 +9,15 @@ def test_jpss_xtce_packet_parsing(jpss_test_data_dir):
...
     assert isinstance(jpss_definition, definitions.XtcePacketDefinition)
 
+    # Prove that parsed objects are referencing the same objects, not duplicates
+    unused_secondary_header_container_ref = jpss_definition.containers["UNUSED"].entry_list[0]
+    assert isinstance(unused_secondary_header_container_ref, containers.SequenceContainer)
+    jpss_att_ephem_header_container_ref = jpss_definition.containers["JPSS_ATT_EPHEM"].entry_list[0]
+    assert isinstance(jpss_att_ephem_header_container_ref, containers.SequenceContainer)
+    assert unused_secondary_header_container_ref in jpss_definition.containers.values()
+    assert jpss_att_ephem_header_container_ref in jpss_definition.containers.values()
+    assert unused_secondary_header_container_ref is jpss_att_ephem_header_container_ref
```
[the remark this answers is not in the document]
medley56: Agreed. This is really a test for a specific regression.

medley56:
@greglucas I profiled creation of the CTIM XTCE definition from XML and it actually got even faster, but that might be the hardware.

```
It took 0.21 seconds to parse the XTCE document
Number of containers: 39
Number of parameters: 9493
Number of parameter types: 15

Process finished with exit code 0
```

greglucas:
Yeah, that is fine with me. Just making sure we didn't go back to the 10+ seconds case. Thanks for profiling!

medley56:
@greglucas All of that added time was due to the checking for unused parameters and parameter types. I removed that checking since it really isn't an error, just an indication that you aren't being very careful with your XTCE. I'm thinking we can instead add that kind of checking to whatever validator we land on out of #129 

I also added a benchmark test for this. We're back down to .03s

space_packet_parser/common.py
```diff
@@ -42,6 +42,7 @@ def from_xml(
             tree: Optional[ElementTree.ElementTree],
             parameter_lookup: Optional[dict[str, any]],
             parameter_type_lookup: Optional[dict[str, any]],
+            container_lookup: Optional[dict[str, any]],
```
greglucas, line 45: Does `any` actually work with mypy, I thought it wanted `typing.Any`.

A few minor nits/questions:
* Do we want to call these "parameters", "parameter_types", "containers" so it is equivalent to what is stored on the definition? (Thinking of this now before we make a major version release)
* Should these be typed as `dict[str, Parameter]`, `dict[str, ParameterType]`, `dict[str, Parameter | SequenceContainer]`
* Since many of these are repeated all over the place, do you want to bring any of these types into a typing helpers / commons import area so you don't have to redefine them all the time?

greglucas (approved):
Looks good to me. I have one minor nit comment that can be addressed in a follow-up PR if we so desire to do any of those things.

space_packet_parser/common.py
medley56: I believe `any` is valid the same way `dict` and `list` are valid. There's also a `typing.Any` similar to `typing.Dict` and `typing.List`.

Let's address this in a future PR.

medley56: ...Before we release a new major version.

greglucas: `any` is a function, not a type though. I've had mypy complain in the past for any (maybe we need a mypy type-checking action too), but perhaps they've changed the typing landscape now.

PR #137, opened 2025-02-12
MNT: Add extras groups for easier installation of optional packages

### LINKED ISSUE #131: Add a test or dev extra for installation
We now have several "optional" dependencies, but have to manually install the packages to test the full suite. It would be nice to have an "extras" for test or dev purposes where we can put these. This would enable `pip install .[test]` to install pytest and all subtests. (Yes, there is poetry-specific development dependencies, but there is no way to get those installed from a different package manager as far as I know)

This is where the extras are defined, with currently only xarray.
https://github.com/medley56/space_packet_parser/blob/d0a608dbcd18abcabb3652a7c74f6052e572079d/pyproject.toml#L65-L66

greglucas:
closes #131

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,16 @@ rich = "^13.0"
 # Extras dependencies
 xarray = { version = ">2024.0.0", optional = true }
 numpy = { version = "^2.0.0", optional = true }
+pytest = { version = "*", optional = true }
+pytest-benchmark = { version = "*", optional = true }
+pytest-randomly = { version = "*", optional = true }
+pytest-cov = { version = "*", optional = true }
+pyyaml = { version = "*", optional = true }
+ruff = { version = "*", optional = true }
+sphinx = { version = "*", optional = true }
+myst-parser = { version = "*", optional = true }
+sphinx-autoapi = { version = "*", optional = true }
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -65,6 +75,9 @@ memory-profiler = "^0.61.0"
 
 [tool.poetry.extras]
 xarray = ["xarray", "numpy"]
+test = ["pytest", "pytest-benchmark", "pytest-randomly", "pytest-cov"]
+dev = ["xarray", "numpy", "pytest", "pytest-benchmark", "pytest-randomly",
+       "pytest-cov", "ruff", "pyyaml", "sphinx", "myst-parser", "sphinx-autoapi", "sphinx-rtd-theme"]
```
greglucas, line 80: Unfortunately poetry doesn't let you self-reference. Other backends let you do `dev = ["space_packet_parser[test,xarray]"` to combine other extras.

medley56: Yeah, I'm falling out of love with Poetry as time goes on... If you want to explore other build backends, I'm open to it. Meson felt a little unfinished to me, at least wrt Rust, but maybe that's a better option and we can just use pip for local development?

medley56 (approved):
Thanks for doing this!

PR #138, opened 2025-02-12
Add pre-commit hooks

greglucas:
This adds pre-commit to the repository. We can optionally switch over to `pre-commit.ci` as well which would be instead of running the linting job as a GitHub Action. Otherwise, this is mostly for myself so that it can check things automatically locally via git hooks.

There are several hooks I added that are nice to have and codespell detected some spelling errors that I fixed up here as well. I also decided to change `smoot` to `m/s` because codespell was complaining about it wanting to be smooth, this is also a more recognizable unit.

One other controversial change that I'd want to make, but will hold off on is adding `ruff format` check as well to auto-format the code and not have to worry about thinking about how to indent etc...

## Checklist


pyproject.toml
```diff
@@ -42,6 +42,16 @@ rich = "^13.0"
...
+pytest = { version = "*", optional = true }
+pytest-benchmark = { version = "*", optional = true }
+pytest-randomly = { version = "*", optional = true }
+pytest-cov = { version = "*", optional = true }
+pyyaml = { version = "*", optional = true }
+ruff = { version = "*", optional = true }
+sphinx = { version = "*", optional = true }
+myst-parser = { version = "*", optional = true }
+sphinx-autoapi = { version = "*", optional = true }
+sphinx-rtd-theme = { version = "*", optional = true }
```
medley56, line 54: I think we also need `pre-commit` as a dev group and dev extra dependency.

medley56 (commented):
Do we also need to update documentation on setting up development environment? i.e. running `pre-commit install`?

medley56:
Branch needs a rebase onto or merge from main, and maybe a doc and dependency update.

pyproject.toml
greglucas: Good call, updated.

medley56 (approved):
Awesome! LGTM

PR #139, opened 2025-02-13
MNT: Move parameter data types to common module

greglucas:
They are not related to packets directly and are used in several different modules for parsing. They could probably go in parameter_types, but that could potentially lead to confusion about whether it is an XTCE StringParameterType or a data value of StrParameter. The generic place is in common, since these are really subclasses of builtins.

## Checklist


medley56 (approved):
Seems reasonable. No comments.

PR #140, opened 2025-02-13
Generic packet

### LINKED ISSUE #130: Add the ability to parse any binary packet from XTCE
Currently we are tied into the CCSDS specification and yielding packets based upon this, but XTCE does not require this. We should try to make it easier to handle other types of packets as well, and I don't think it would be too difficult, but would involve some book keeping and renaming of classes.

We could keep our `RawPacketData` class with the read_* methods on it as a base class.
https://github.com/medley56/space_packet_parser/blob/d0a608dbcd18abcabb3652a7c74f6052e572079d/space_packet_parser/packets.py#L84
Then subclass this to add on all of the CCSDS header parsing stuff with a new `RawCCSDSPacketData` class. If there are other packet types, we could keep adding subclasses to handle different headers etc, but leave the reader up on the superclass.

Create a generic `Packet` class that doesn't have the header/user_data fields, very plain to store parsed content from XTCE. Again subclassing dict and allowing for our current `CCSDSPacket` to add those methods as a subclass.
https://github.com/medley56/space_packet_parser/blob/d0a608dbcd18abcabb3652a7c74f6052e572079d/space_packet_parser/packets.py#L278

**XTCE Definition**

Rather than specific `parse_ccsds_packet` methods, we could call this `parse_packet()` and allow for any byte representation to be passed in.
https://github.com/medley56/space_packet_parser/blob/d0a608dbcd18abcabb3652a7c74f6052e572079d/space_packet_parser/definitions.py#L312

The packet generator would be a bit trickier and what to do with that, because it is very much tied to a user's chosen spec and needs to know how many bytes to pull off at any given time.
https://github.com/medley56/space_packet_parser/blob/d0a608dbcd18abcabb3652a7c74f6052e572079d/space_packet_parser/definitions.py#L362

@cgobat, you mentioned you had some other formats of packets besides CCSDS that would be useful to parse, so it'd be good to hear your thoughts on what would be helpful to make this more generic and if the above seems reasonable or there are better ways to approach this.

greglucas:
This is an attempt to refactor our xtce parsing to be more generic and allow any RawDataObject, not just ccsds packets.

* RawPacketData has the read_* methods, RawCCSDSPacketData subclasses that and adds on cached properties to access specific header items.
* Moved the ccsds continuation packet logic into the ccsds_generator to try and keep all ccsds logic in one place.

Thoughts/suggestions on how to make this better for other packet types are welcome!
* I was thinking it might be nice if a user could pass in raw bytes directly to `definition.parse_packet(bytes)` and not our wrapped classes that add the reading methods. Perhaps we could wrap the input bytes into our own container class within the function itself which would remove some of the user-facing need to create a RawPacketData-like object with read_* methods on it.

closes #130 

## Checklist


greglucas:
I've updated this to add new methods / names for generic methods (avoiding ccsds), but I tried to keep backwards compatibility and add warnings to the old methods.

packet_generator is still an oddity that I don't know how we want to handle. It is currently using the ccsds_generator, and it needs some form of generator to yield raw bytes... We could say it is a "bring your own generator" method defaulting to ccsds_packet_generator, but that would also require passing extra kwargs etc. along and becomes a bit nasty.

medley56:
@greglucas Can you rebase this? Then I'll review.

space_packet_parser/packets.py
```diff
@@ -29,13 +30,67 @@ class SequenceFlags(IntEnum):
...
+        """
+        if self.pos + nbits > len(self) * 8:
+            raise ValueError("Tried to read beyond the end of the packet data. "
+                             f"Tried to read {nbits} bits from position {self.pos} "
+                             f"in a packet of length {len(self) * 8} bits.")
+        int_data = _extract_bits(self, self.pos, nbits)
+        self.pos += nbits
+        return int_data
+
+class CCSDSPacketBytes(bytes):
```
medley56, line 90: Intuitively, I would expect `CCSDSPacketBytes` to subclass `RawPacketData` and to be called `CCSDSRawPacketData` to make it clear that they are related. Perhaps I misunderstand but both `RawPacketData` and `CCSDSPacketBytes` are just wrappers around `bytes`, but every method for `RawPacketData` should also be applicable to `CCSDSPacketBytes`, right?

Maybe more simply, `RawPacketData(bytes)` -> `PacketData(bytes)` (data meaning bytes), and `CCSDSPacketBytes(bytes)` -> `CCSDSPacketData(PacketData)`?

`CCSDSPacketData` would be strictly Liskov in that every `CCSDSPacketData` is a valid `PacketData` object because we're not adding limitations, only functionality (for parsing header fields as properties).

space_packet_parser/packets.py
```diff
@@ -29,13 +30,67 @@ class SequenceFlags(IntEnum):
...
+    """Binary representation of a CCSDS packet.
+
+    Methods to extract the header fields are added to the raw bytes object.
```
medley56, line 93: The `__str__` dunder is still calling this `RawPacketData`. Probably meant to change that.

space_packet_parser/packets.py
```diff
@@ -29,13 +30,67 @@ class SequenceFlags(IntEnum):
...
-    """A class to represent raw packet data as bytes but whose length is represented by bit length.
+    """A class containing the raw binary packet data.
 
     This class is a subclass of bytes and is used to represent the raw packet data
-    in a more readable way. It is used to store the raw packet data in the Packet
+    in a more readable way. It is used to store the binary data in the Packet
     class and used to keep track of the current parsing position (accessible through the `pos` attribute).
+    """
+    pos = 0  # Current position in bits
+
```
medley56, line 40: Did you want to give `RawPacketData` a `__str__` dunder like `CCSDSPacketBytes`? If we follow the inheritance suggestion below, we could just define it once in the (this) parent class.

space_packet_parser/packets.py
```diff
@@ -245,21 +256,47 @@ def __init__(self, *args, raw_data: bytes = b"", **kwargs):
...
+    Container that stores the raw packet data (bytes) as an instance attribute and the parsed
+    data items in a dictionary interface. A ``CCSDSPacket`` generally begins as an empty dictionary that gets
+    filled as the packet is parsed. The first 7 items in the dictionary make up the
+    packet header (accessed with ``CCSDSPacket.header``), and the rest of the items
+    make up the user data (accessed with ``CCSDSPacket.user_data``). To access the
+    raw bytes of the packet, use the ``CCSDSPacket.raw_data`` attribute.
+    """
+    def __init__(self, *args, **kwargs):
+        warnings.warn("The CCSDSPacket class is deprecated and will be removed in a future release. "
```
medley56, line 286: Make this an explicit `DeprecationWarning`. See https://docs.python.org/3/library/warnings.html#warnings.warn
The `stacklevel` kwarg can kick the warning up the stack so that it refers the user to the code that is _calling_ the deprecated code rather than pointing to the deprecated code itself.

space_packet_parser/packets.py
```diff
@@ -245,21 +256,47 @@ def __init__(self, *args, raw_data: bytes = b"", **kwargs):
     @property
     def header(self) -> dict:
         """The header content of the packet."""
+        warnings.warn("The header property is deprecated and will be removed in a future release. "
```
medley56, line 259: DeprecationWarning

space_packet_parser/packets.py
```diff
@@ -245,21 +256,47 @@ def __init__(self, *args, raw_data: bytes = b"", **kwargs):
...
+                      "To access the header fields of a CCSDS packet, use the CCSDSPacketBytes class.")
         return dict(list(self.items())[:7])
 
     @property
     def user_data(self) -> dict:
         """The user data content of the packet."""
+        warnings.warn("The user_data property is deprecated and will be removed in a future release. "
```
medley56, line 266: DeprecationWarning

space_packet_parser/packets.py
```diff
@@ -245,21 +256,47 @@ def __init__(self, *args, raw_data: bytes = b"", **kwargs):
...
             buffer_read_size_bytes: Optional[int] = None,
             show_progress: bool = False,
             skip_header_bytes: int = 0,
-) -> Iterator[RawPacketData]:
+            combine_segmented_packets: bool = False,
+            secondary_header_bytes: int = 0,
+) -> Iterator[CCSDSPacketBytes]:
     """A generator that reads raw packet data from a filelike object or a socket.
 
     Each iteration of the generator yields a ``RawPacketData`` object that makes up
```
medley56, line 302: ...yields a `CCSDSPacketBytes` object...
But if we made it inherit from RawPacketData, then this would technically be true.

space_packet_parser/packets.py
```diff
@@ -29,13 +30,67 @@ class SequenceFlags(IntEnum):
...
+        self.pos += nbits
+        return int_data
+
+class CCSDSPacketBytes(bytes):
+    """Binary representation of a CCSDS packet.
+
+    Methods to extract the header fields are added to the raw bytes object.
     """
     HEADER_LENGTH_BYTES = 6
     pos = 0  # in bits
```
medley56, line 96: Would it be appropriate to add bookkeeping to the `CCSDSPacketBytes` class to keep track of the necessary information for concatenating and parsing continued packets? The signature of the class initializer wouldn't need to change but we could add instance attributes that mark it as a continued packet, how many packets are continued, and how big the secondary header is. The idea would be to self contain all the necessary information needed to correctly parse it later.

space_packet_parser/packets.py
```diff
@@ -369,7 +420,37 @@ def ccsds_generator(
...
-        yield RawPacketData(packet_bytes)
+        ccsds_packet = CCSDSPacketBytes(packet_bytes)
+
+        if not combine_segmented_packets or ccsds_packet.sequence_flags == SequenceFlags.UNSEGMENTED:
+            yield ccsds_packet
+        elif ccsds_packet.sequence_flags == SequenceFlags.FIRST:
+            _segmented_packets[ccsds_packet.apid] = [ccsds_packet]
+            continue
+        elif not _segmented_packets.get(ccsds_packet.apid, []):
+            warnings.warn("Continuation packet found without declaring the start, skipping this packet.")
```
medley56, line 431: Log the APID in this warning

space_packet_parser/packets.py
```diff
@@ -369,7 +420,37 @@ def ccsds_generator(
...
+            continue
+        else:  # raw_packet_data.sequence_flags == packets.SequenceFlags.LAST:
+            _segmented_packets[ccsds_packet.apid].append(ccsds_packet)
+            # We have received the final packet, close it up and combine all of
+            # the segmented packets into a single "packet" for XTCE parsing
+            sequence_counts = [p.sequence_count for p in _segmented_packets[ccsds_packet.apid]]
+            if not all((sequence_counts[i + 1] - sequence_counts[i]) % 16384 == 1
+                        for i in range(len(sequence_counts) - 1)):
+                warnings.warn(f"Continuation packets for apid {ccsds_packet.apid} "
+                              f"are not in sequence {sequence_counts}, skipping these packets.")
```
medley56, line 444: I think it would be safe to just reorder these according the sequence counter, but maybe leave the warning even if we can fix it because it's probably not expected that the packets are out of order.

space_packet_parser/packets.py
```diff
@@ -369,7 +420,37 @@ def ccsds_generator(
...
+                continue
+            # Add all content (including header) from the first packet
+            raw_data = _segmented_packets[ccsds_packet.apid][0]
+            # Add the continuation packets to the first packet, skipping the headers
+            for p in _segmented_packets[ccsds_packet.apid][1:]:
+                raw_data += p[header_length_bytes + secondary_header_bytes:]
```
medley56, line 450: It still bothers me that we're simply losing headers and secondary headers by doing this. Could we store the ignored headers and secondary headers as a `list` instance attribute on the `CCSDSPacketBytes` object that we yield? See my comment above about bookkeeping this information.

space_packet_parser/packets.py
```diff
@@ -369,7 +420,37 @@ def ccsds_generator(
...
+            yield CCSDSPacketBytes(raw_data)
```
medley56, line 451: I think we have a minor memory leak in `_segmented_packets`. We should `pop` the APID key of the `_segmented_packets` dict for this packet in order to yield the combined `CCSDSPacketBytes` object.

Upon further inspection, this probably isn't a huge deal because the biggest this dictionary will ever be is the number of distinct muxed APIDs in the packet stream being parsed. It still feels cleaner to me to pop the continued packet from the dict rather than just leaving it there.

space_packet_parser/xtce/definitions.py
```diff
@@ -371,15 +371,38 @@ def _parse_parameter_set(
 
         return parameter_lookup
 
-    def parse_ccsds_packet(self,
-                           packet: packets.CCSDSPacket,
-                           *,
-                           root_container_name: Optional[str] = None) -> packets.CCSDSPacket:
+    def parse_bytes(self,
```
medley56, line 374: We're not just parsing any bytes object though. We're parsing bytes of packet data. I'd suggest naming this `parse_packet_data`. I'm noticing that I the idiom where "packet data" is always a bytes object and the "Packet" is a dict.

space_packet_parser/xtce/definitions.py
```diff
@@ -434,7 +482,7 @@ def packet_generator(
             show_progress: bool = False,
             buffer_read_size_bytes: Optional[int] = None,
             skip_header_bytes: int = 0
-    ) -> Iterator[Union[packets.CCSDSPacket, UnrecognizedPacketTypeError]]:
+    ) -> Iterator[Union[packets.Packet, UnrecognizedPacketTypeError]]:
```
medley56, line 485: So if we had a non-CCSDS packet, would we need a separate generator method (and associated generator function)? This method _sounds_ agnostic to CCSDS packets (`packet_generator()`) but internally it only works for CCSDS packets. 

I'm not sure how to handle that better. i.e. I'm not sure how to decide whether to call the CCSDS generator or the non-CCSDS generator. Do we want the user telling us whether they are using CCSDS packets or not when initializing the packet definition object? That would make it awkward to support mixed definitions where some are CCSDS and others are not.

medley56 (changes requested):
I tried to be really thorough. I had a few suggestions on semantics and a significant change in inheritance structure. One big question about how to make the definition-level `packet_generator()` agnostic to CCSDS.

space_packet_parser/xtce/definitions.py
greglucas: This is related to some of the other comments about inheritance, but this is just any bytes object actually! That is the point of this new function, we want to allow `bytes` as input, whatever a user wants to pass to us bytes-like.

Once we get the `bytes`, we then wrap that in our own class that has the reader methods on it. This is why I didn't do the full inheritance from `RawPacketData` elsewhere on `CCSDSPacketBytes`. I want a user to be able to make their own bytes generator, I don't want them to have to make a `RawPacketData(bytes)` generator if that makes sense. Maybe there is a better way to do this though? I don't love that we are wrapping a bytes object in another bytes-like object, but this was the quickest I came up with.

space_packet_parser/xtce/definitions.py
greglucas: Yes, this is definitely CCSDS specific, because that is where we started from. We could do something like "bring your own generator" as an input argument. But, this sort of gets at a definition probably shouldn't have the generator on it, the generator should be separate and then each yielded item can be parsed by the `definition.parse_bytes()` method. Or we would grow a separate generator for each yielded possibility.

space_packet_parser/packets.py
greglucas: This accounts for gaps though too, so it isn't just re-ordering.

space_packet_parser/packets.py
greglucas: I'm not sure I follow what you mean by correctly parse it later? The XTCE definition should not contain any information about the inbetween headers, so those bytes are meant to be dropped in the CCSDS spec.

If a user wants that information then they should call this function without concatenating the data together I think. I don't see it being super useful personally to bookkeep that data because there is a way to get it by going back to the source.

space_packet_parser/packets.py
greglucas: Yeah, I generally don't use DeprecationWarning because it is hidden by default unfortunately. I can add the stacklevel here to boost them.

space_packet_parser/packets.py
greglucas: Said in the other comment, but I don't think that the `CCSDSPacketBytes` should be "readable". It is just a stream of bytes, and there are some extra methods that do some nice things on the header bits for ease of access. The reader stuff within `RawPacketData` is really only useful for XTCE-parsing things. This is external to that and someone can just iterate over the raw bytes to count packets etc.. without any reading (like we are doing in some of the cli stuff).

Suggestions for improving this are welcome, but I don't think we should require someone to bring a `RawPacketData` subclass to the definition parsing, we should help them out and add the readers ourselves since that is really only meant for us.

space_packet_parser/packets.py
greglucas: Every time we get a new START packet, we create a new storage list. But, I think you're right that this could grow if we get something like LAST, CONTINUATION, CONTINUATION, LAST, CONTINUATION, because there would be packets in the queue still because we never got another START. I added the pop in because I think that is a good way to handle this, thanks for the suggestion!

space_packet_parser/common.py
```diff
@@ -247,3 +246,144 @@ class StrParameter(_Parameter, str):
...
+            binary_data = kwargs.pop("raw_data")
+        self.binary_data = binary_data
+        self._parsing_pos = 0
+        super().__init__(*args, **kwargs)
+
+    @property
+    def header(self) -> dict:
+        """The header content of the packet."""
+        warnings.warn("The header property is deprecated and will be removed in a future release. "
+                      "To access the header fields of a CCSDS packet, use the CCSDSPacketBytes class.",
```
medley56, line 282: Is the intention that users should be creating `Packets` where the `binary_data` is a `CCSDSPacketBytes` object rather than a plain `bytes` object? Then they would call `my_packet.binary_data.header`? I think we should be more explicit about how to do this and maybe even explain that this is to support generic, non CCSDS packets. Also this applies to the warning in the `user_data` property.

space_packet_parser/xtce/definitions.py
medley56: I see how this is generic now (not CCSDS only) but why do we have the layer of abstraction of `parse_bytes` calling `parse_packet`? I don't see `parse_packet` being used anywhere else. Is there a use case for `parse_packet` that is not being called by `parse_bytes`?

space_packet_parser/xtce/definitions.py
medley56: Per your Slack message, I think you suggested we make this generator part of the definition but I'm not sure I fully understand how that would work.

space_packet_parser/xtce/definitions.py
```diff
@@ -455,17 +503,8 @@ def packet_generator(
             point for parsing. Default is taken from the parent XtcePacketDefinition object.
         ccsds_headers_only : bool
             Default False. If True, only parses the packet headers (does not use the provided packet definition).
-            ``space_packet_parser.packets.ccsds_packet_generator`` can be used directly to parse only the CCSDS headers
+            ``space_packet_parser.ccsds.ccsds_packet_generator`` can be used directly to parse only the CCSDS headers
```
medley56, line 506: Should the `ccsds_headers_only` kwarg be deprecated since `packet_generator` is supposed to be generic?

space_packet_parser/xtce/definitions.py
```diff
@@ -493,64 +532,30 @@ def packet_generator(
...
-        # for the XTCE to parse, lookup is by APID.
-        # _segmented_packets[APID] = [RawPacketData, ...]
-        _segmented_packets = {}
-
         # Iterate over individual packets in the binary data
-        for raw_packet_data in packets.ccsds_generator(binary_data,
-                                                       buffer_read_size_bytes=buffer_read_size_bytes,
-                                                       show_progress=show_progress,
-                                                       skip_header_bytes=skip_header_bytes):
+        for raw_packet_data in ccsds.ccsds_generator(binary_data,
```
medley56, line 536: Should this generator be default kwarg to `packet_generator` so users can pass custom non-CCSDS generators? I know you have a vision for this but this feels like we're only halfway to being truly generic because the generic `packet_generator` is hard-coded to use the CCSDS-specific bytes generator here.

medley56 (commented):
I still am not convinced the packet_generator method on the definition object is truly generic because it's calling `ccsds_generator` internally to create the `CCSDSPacketBytes` objects for parsing. The _parsing_ appears to be fully generic but the generation of the packet bytes seems to still rely on CCSDS.

space_packet_parser/xtce/encodings.py
```diff
@@ -99,12 +100,12 @@ def adjuster(x: float) -> float:
...
         return None
 
-    def _calculate_size(self, packet: packets.CCSDSPacket) -> int:
+    def _calculate_size(self, packet: spp.SpacePacket) -> int:
         """Calculate the size of the data item in bits.
 
         Parameters
         ----------
-        packet: CCSDSPacket
+        packet: Packet
```
medley56, line 108:
```suggestion
        packet: SpacePacket
```

space_packet_parser/xtce/encodings.py
```diff
@@ -115,12 +116,12 @@ def _calculate_size(self, packet: packets.CCSDSPacket) -> int:
...
         return NotImplemented
 
-    def parse_value(self, packet: packets.CCSDSPacket) -> common.ParameterDataTypes:
+    def parse_value(self, packet: spp.SpacePacket) -> common.ParameterDataTypes:
         """Parse a value from packet data, possibly using previously parsed data items to inform parsing.
 
         Parameters
         ----------
-        packet: CCSDSPacket
+        packet: Packet
```
medley56, line 124:
```suggestion
        packet: SpacePacket
```

space_packet_parser/xtce/encodings.py
```diff
@@ -237,12 +238,12 @@ def __init__(
...
         self.length_linear_adjuster = length_linear_adjuster
 
-    def _calculate_size(self, packet: packets.CCSDSPacket) -> int:
+    def _calculate_size(self, packet: spp.SpacePacket) -> int:
         """Calculate the size of the raw string buffer field
 
         Parameters
         ----------
-        packet : packets.CCSDSPacket
+        packet : Packet
```
medley56, line 246:
```suggestion
        packet : SpacePacket
```

medley56: Ok there's a lot of these. I'll leave it to you to fix those typehints

medley56 (approved):
I get it now. I noted a few really minor things. There are a bunch of typehints that read Packet instead of SpacePacket. The example and associated test you added for fix length packet parsing for non CCSDS packets is what I needed to see. LGTM

space_packet_parser/xtce/encodings.py
greglucas: Yes, sorry I guess I was only doing my find+replace over code before and missed the docstring types. Thanks for pointing it out!

PR #141, opened 2025-02-14
CI: Explicitly output the codecov xml file

greglucas:
Seeing if this explicit output option helps to get coverage sent up. Thinking it may be an empty coverage report 🤷 . Just another shot in the dark here...


greglucas:
Closing because codecov is working on main now.

PR #143, opened 2025-02-15
Handle all namespace cases - default, prefixed, no namespace.

medley56:
- Avoid passing namespace dicts down through all from_xml methods in favor of predefining a custom Element class that contains the namespace
- Convert internal parameter types, parameters, and containers to set objects with read only dict accessor properties.
- Increase performance by minimizing broad XML find and findall calls.

## Checklist

### DIFF
--- a/space_packet_parser/common.py
+++ b/space_packet_parser/common.py
@@ -9,6 +9,110 @@ from lxml.builder import ElementMaker
 from space_packet_parser import packets
 
 
+class NamespaceAwareElement(ElementTree.ElementBase):
+    """Custom element that automatically applies namespace mappings."""
+
+    _nsmap: dict[str, str] = {}  # Class level namespace mapping
+    _ns_prefix: Union[str, None] = None  # Class level namespace prefix for adding to Xpaths
+
+    @classmethod
+    def element_prefix(cls):
+        """Create the XPath element prefix
+
+        Notes
+        -----
+        If the prefix is None,
+        it indicates either an implicit namespace such as `<SpaceSystem xmlns="http://xtce-example-ns-uri"/>`,
+        where `nsmap` is `{None: "http://xtce-example-ns-uri", ...}`
+        or no namespace awareness, such as `<SpaceSystem/>`,
+        where `nsmap` does not contain any reference to a URI or prefix for the XTCE namespace
+        (but may contain other namespace mappings).
+
+        If the prefix is anything other than None,
+        it must be a string and must be present in the namespace mapping dict and represents a prefixed namespace,
+        such as `<xtce:SpaceSystem xmlns:xtce="http://xtce-example-ns-uri"/>`
+        where `nsmap` would be `{"xtce": "http://xtce-example-ns-uri", ...}` and `ns_prefix` would be `xtce`.
+        """
+        if cls._ns_prefix is not None:
+            if cls._ns_prefix not in cls._nsmap:
+                raise ValueError(f"XTCE namespace prefix {cls._ns_prefix} not found in namespace mapping "
+                                 f"{cls._nsmap}. If the namespace prefix is not 'None', it must appear as a key in the "
+                                 f"namespace mapping dict.")
+            return f"{cls._ns_prefix}:"
+        return ""
+
+    @classmethod
+    def add_namespace_to_xpath(cls, xpath: str) -> str:
+        """
+        Adds a namespace prefix to each element in an XPath expression.
+
+        Parameters
+        ----------
+        xpath : str
+            The original XPath expression without namespace prefixes.
+
+        Returns
+        -------
+        str
+            The updated XPath expression with namespace prefixes.
+        """
+        prefix = cls.element_prefix()
+        # Regex to match valid XML element names (avoids matching special characters like `@attr`, `.`, `*`, `()`, `::`)
+        xpath_parts = xpath.split('/')
+        new_parts = []
+
+        for part in xpath_parts:
+            # Skip empty parts (handles leading/trailing slashes)
+            if not part:
+                new_parts.append('')
+                continue
+
+            # Handle special cases (wildcards, functions, attributes, self, parent, axes)
+            if part is None or part in {'.', '..', '*'} or part.startswith('@') or '::' in part or '(' in part:
+                new_parts.append(part)
+            else:
+                new_parts.append(f"{prefix}{part}")
+
+        new_path = '/'.join(new_parts)
+        return new_path
+
+    def find(self, path, namespaces=None):
+        """Override find() to automatically use the stored namespace map."""
+        if namespaces is None:
+            namespaces = self.get_nsmap()
+        return super().find(self.add_namespace_to_xpath(path), namespaces=namespaces)
+
+    def findall(self, path, namespaces=None):
+        """Override findall() to automatically use the stored namespace map."""
+        if namespaces is None:
+            namespaces = self.get_nsmap()
+        return super().findall(self.add_namespace_to_xpath(path), namespaces=namespaces)
+
+    def iterfind(self, path, namespaces=None):
+        """Override iterfind() to automatically use the stored namespace map."""
+        if namespaces is None:
+            namespaces = self.get_nsmap()
+        return super().iterfind(self.add_namespace_to_xpath(path), namespaces=namespaces)
+
+    @classmethod
+    def set_nsmap(cls, nsmap: dict):
+        """Store the namespace map for all elements of this type."""
+        cls._nsmap = nsmap
+
+    def get_nsmap(self):
+        """Retrieve the stored namespace map."""
+        return self._nsmap
+
+    @classmethod
+    def set_ns_prefix(cls, ns_prefix: Union[str, None]):
+        """Store the namespace map for all elements of this type."""
+        cls._ns_prefix = ns_prefix
+
+    def get_ns_prefix(self):
+        """Retrieve the stored namespace map."""
+        return self._ns_prefix
+
+
 # Common comparable mixin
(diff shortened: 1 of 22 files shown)


space_packet_parser/xtce/definitions.py
```diff
@@ -40,82 +46,81 @@ class XtcePacketDefinition(common.AttrComparable):
     #  This will require an additional namespace dict entry for xsi (in this example)
     def __init__(
             self,
-            container_list: Optional[list[containers.SequenceContainer]] = None,
+            container_set: Optional[set[containers.SequenceContainer]] = None,
```
greglucas, line 49: I think we want to allow any iterable thing. Maybe indicate that it should be unique or we warn if it isn't.
```suggestion
            container_set: Optional[Iterable[containers.SequenceContainer]] = None,
```

space_packet_parser/xtce/definitions.py
```diff
@@ -40,82 +46,81 @@ class XtcePacketDefinition(common.AttrComparable):
...
             Name of space system to encode in XML when serializing.
+        validation_status : str
+            One of ["Unknown", "Working", "Draft", "Test", "Validated", "Released", "Withdrawn"].
         xtce_version : str
             Default "1.0"
         date: Optional[str]
             Optional header date string.
         """
+        if xtce_ns_prefix and xtce_ns_prefix not in ns:
+            raise ValueError(f"XTCE namespace prefix {xtce_ns_prefix} not in namespace mapping {ns}. This is an error.")
```
greglucas, line 85: Let me know how to fix this error in the message.

space_packet_parser/xtce/definitions.py
```diff
@@ -40,82 +46,81 @@ class XtcePacketDefinition(common.AttrComparable):
...
+            ----------
+            sc : containers.SequenceContainer
+                The SequenceContainer to iterate through.
+            """
             self.containers[sc.name] = sc
             for entry in sc.entry_list:
                 if isinstance(entry, containers.SequenceContainer):
                     _update_caches(entry)  # recurse
                 elif isinstance(entry, parameters.Parameter):
-                    if entry.name in self.parameters and self.parameters[entry.name] is not entry:
```
greglucas, line 91: This is just overwriting the previous object with the new one now it looks like. Do you want to warn/log instead of raising still?

space_packet_parser/xtce/parameter_types.py
```diff
@@ -33,6 +33,9 @@ def __init__(self, name: str, encoding: encodings.DataEncoding, unit: Optional[s
         self.encoding = encoding
         self.unit = unit
 
+    def __hash__(self) -> int:
+        return hash(self.name)
```
greglucas, line 37: Do we need these hashes redefined? The keys in the dictionary lookups are still string names, so the dictionary hash is this in a sense. i.e. I'm not sure we are using the object hashes anywhere?

tests/integration/test_xtce_based_parsing/test_inheritance_restrictions.py
```diff
@@ -9,6 +9,15 @@ def test_jpss_xtce_packet_parsing(jpss_test_data_dir):
     jpss_definition = definitions.XtcePacketDefinition.from_xtce(xtce_document=jpss_xtce)
     assert isinstance(jpss_definition, definitions.XtcePacketDefinition)
 
+    # Prove that parsed objects are referencing the same objects, not duplicates
```
greglucas, line 12: I thought you had this in your previous PR?

tests/unit/test_xtce/test_comparisons.py
```diff
@@ -4,172 +4,171 @@
...
+from space_packet_parser.xtce import XTCE_1_2_XMLNS, comparisons
 
 
 @pytest.mark.parametrize(
     ('xml_string', 'test_parsed_data', 'current_parsed_value', 'expected_comparison_result'),
     [
-        ("""
-<xtce:Comparison xmlns:xtce="http://www.omg.org/space/xtce"
+        (f"""
+<xtce:Comparison xmlns:xtce="{XTCE_1_2_XMLNS}"
```
greglucas, line 14: Do you need the namespace defined here, or can we just remove it altogether? Either `<Comparison>` or `<xtce:Comparison>` without the trailing text.

greglucas (commented):
I don't know enough about namespaces or XML, so most of my comments are just my ignorance of it all. Feel free to dismiss or ignore. Overall, I really like that this gets rid of the `ns` showing up everywhere and think it will simplify things moving forward!

space_packet_parser/common.py
```diff
@@ -9,6 +9,94 @@
 from space_packet_parser import packets
 
 
+class NamespaceAwareElement(ElementTree.ElementBase):
+    """Custom element that automatically applies namespace mappings."""
+
+    _nsmap = {}  # Class level namespace mapping
+    _ns_prefix = None  # Class level namespace prefix for adding to Xpaths
```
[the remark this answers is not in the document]
medley56: I think either would work with the current implementation of our custom Element class. Basically there are a few different cases:
1. Namespaced with a prefix, e.g. `<prefixstr:SpaceSystem xmlns:prefixstr="http://ns-uri.com">`, where `nsmap={"prefixstr": "http://ns-uri.com"}` and `xtce_ns_prefix="prefixstr"`
2. Namespaced but without an explicit prefix, e.g. `<SpaceSystem xmlns="http://ns-uri.com">` where `nsmap={None: "http://ns-uri.com"}` (this is how lxml parses a namespaced tree without a prefix) and `xtce_ns_prefix=None`.
3. No namespace awareness, e.g. `<SpaceSystem>`, where `nsmap={}` and `xtce_ns_prefix=None`.

Note that `nsmap` is produced by `lxml` when it parses an XML tree from a string or file. We don't control what that looks like. But we need to know which namespace is associated with XTCE so that we can use it later to parse/serialize XML from objects, which is why we need to know `xtce_ns_prefix`.

space_packet_parser/common.py
```diff
@@ -9,6 +9,94 @@
...
+        xpath : str
+            The original XPath expression without namespace prefixes.
+
+        Returns
+        -------
+        str
+            The updated XPath expression with namespace prefixes.
+        """
+        prefix = cls.element_prefix()
+        # Regex to match valid XML element names (avoids matching special characters like `@attr`, `.`, `*`, `()`, `::`)
```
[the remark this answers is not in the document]
medley56: There might be better ways to do this, but this is what I found in the `lxml` docs (combined with some GPT help).
https://lxml.de/element_classes.html
https://lxml.de/element_classes.html#implementing-namespaces-1

cgobat:
I haven't reviewed the code in detail, but I just want to say I think this is a great idea. It has always baffled me a bit that with lxml, something as simple as a `.find()` call has to become `my_element.find("my_prefix:other_tag", namespaces=my_element.nsmap)`. It makes a lot more sense for `my_element` to just be self-aware of its own namespace(s).

space_packet_parser/xtce/definitions.py
medley56: I opted to just remove those checks entirely. I do want it to be the expectation but those lookups were eating up a lot of time in the XtcePacketDefinition creation benchmark. 

I think I'd rather have it encapsulated in a validation method that can be turned off if speed is important. Maybe I'll take a stab at a basic validation method for object references.

space_packet_parser/xtce/parameter_types.py
medley56: Oh, thanks for catching that. I had these in when I was planning to implement internal `set` objects but I ditched that idea when I saw the resulting performance hit for doing lookups.

medley56: No hash method does mean that we cannot create set objects from these classes but that's not really necessary currently.

tests/integration/test_xtce_based_parsing/test_inheritance_restrictions.py
medley56: Rebasing mistake. These assertions are in the unit tests for definitions.py. This just snuck back in when I rebased.

tests/unit/test_xtce/test_comparisons.py
medley56: The tests use a standard fixture now that generates an `xtce_parser` that is namespace aware, but since it's using the same fixture for all tests, all tests need to use the same namespacing. We could remove it from _all_ the tests that use the `xtce_parser` fixture...

tests/conftest.py
```diff
@@ -3,15 +3,29 @@
...
+@pytest.fixture
+def xtce_parser():
+    """Parser for testing that knows about the standard testing namespace we use"""
+    el = common.NamespaceAwareElement
+    el.set_nsmap(DEFAULT_XTCE_NSMAP)
+    el.set_ns_prefix(DEFAULT_XTCE_NS_PREFIX)
+    xtce_element_lookup = etree.ElementDefaultClassLookup(element=el)
+    xtce_parser = etree.XMLParser()
+    xtce_parser.set_element_class_lookup(xtce_element_lookup)
+    return xtce_parser
```
medley56, line 28: @greglucas These are the fixtures that assume the `XTCE_1_2_XMLNS` namespace URI and the default namespace mapping.

greglucas (approved):
To me, this looks super helpful and is much appreciated to remove all the required "ns" passing. Thank you for looking into this. All of my minor code-formatting comments have been addressed. I think the logic seems sound, but I don't have enough experience to fully verify that is the best approach. I think we should get this in sooner rather than later due to the amount of code it touches by just removing "ns=" in so many places. We can iterate on the little things later because it is mostly non-user facing.

PR #144, opened 2025-02-15
CI: Remove explicit linting job and use precommit.ci instead

greglucas:
The current merge to main failed because it ran "no commit to a branch named main". I enabled pre-commit.ci service so that will do the linting checks for us now as a separate action. (we could remove that check to fix this too)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,19 +9,6 @@ on:
     - cron: '50 23 * * *'  # 11:50pm daily (to get ahead of all those midnight jobs)
 
 jobs:
-
-  linting:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check out repo
-        uses: actions/checkout@v4
-      - name: Install pre-commit
-        run: pip install pre-commit
-      - name: Run pre-commit
-        run: |
-          pre-commit install
-          pre-commit run --all
-
   ci-tests:
     runs-on: ${{ matrix.os }}
     permissions:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,7 @@
 ci:
   autofix_prs: false
   autoupdate_schedule: 'quarterly'
+  skip: [no-commit-to-branch]
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,6 @@
 ci:
   autofix_prs: false
   autoupdate_schedule: 'quarterly'
-  skip: [poetry-lock]
 repos:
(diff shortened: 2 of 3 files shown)


medley56:
The ReadTheDocs failure is due to a Poetry regression in v2.1.0, released this morning. In my branch I pinned the poetry version in `readthedocs.yml` to 2.0.1.

medley56:
And it looks like our `pre-commit-config.yaml` file has a `skip` reference to a hook called `poetry install` that doesn't actually exist.

greglucas:
Thanks! Pushed a new commit with those fixes.

medley56:
pre-commit.ci run

.github/workflows/tests.yml
```diff
@@ -9,19 +9,6 @@ on:
...
-    steps:
-      - name: Check out repo
-        uses: actions/checkout@v4
-      - name: Install pre-commit
-        run: pip install pre-commit
-      - name: Run pre-commit
-        run: |
-          pre-commit install
-          pre-commit run --all
-
```
medley56, line 24: So this linting job is now handled only by `pre-commit.ci`, which runs whenever PRs are opened or modified. But we are losing the periodic linting of the main branch because the `no-commit-to-branch` hook was failing when run against main. Is that right?

medley56 (approved):

.github/workflows/tests.yml
greglucas: Yes, I believe that is correct. A few options:
1. Don't worry about linting main periodically (current situation on this PR), pushes should have gone through a PR already because of branch protection rules.
2. Remove the no-commit-to-main check since that is the only one failing.
3. Instead of using the pre-commit.ci, use an explicit action/workflow: https://github.com/pre-commit/action (would still need to figure out how to skip (or remove) the commit-to-main check.

.pre-commit-config.yaml
```diff
@@ -1,7 +1,6 @@
 ci:
   autofix_prs: false
   autoupdate_schedule: 'quarterly'
```
medley56, line 3:
```suggestion
  autoupdate_schedule: 'quarterly'
  skip: [no-commit-to-branch]
```

.github/workflows/tests.yml
medley56: Does no-commit-to-main even mean anything in the pre-commit.ci context? That commit has already been made and pushed if pre-commit.ci is running. If I understand correctly, the real utility of "no-commit-to-main" is to catch you if you accidentally forget to switch to a feature branch before you start working and you accidentally try to commit to `main` locally. I think I'm ok with skipping it for CI testing.

.pre-commit-config.yaml
greglucas: Good call, not sure why I didn't think of this when removing the other skip we had. This is much better and yes it is for local developers.

medley56 (approved):

PR #145, opened 2025-02-18
MNT: Add backcompat shim to ease transition to new packet file loading

greglucas:
When I was trying to profile things across versions, it got a little annoying to comment/uncomment lines of files for the different loading XTCE patterns. This adds a small shim layer for the old `XtcePacketDefinition` format that then warns and suggests how to transition to the new patterns, still working the same as previously.

### DIFF
--- a/space_packet_parser/definitions.py
+++ b/space_packet_parser/definitions.py
@@ -1,0 +1,17 @@
+# NOTE: This entire file is deprecated and here just to maintain backwards compatibility.
+# The functionality has been moved to space_packet_parser.xtce.definitions.XtcePacketDefinition
+import warnings
+
+from space_packet_parser.xtce import definitions
+
+warnings.warn("The space_packet_parser.definitions module is deprecated. "
+              "Use space_packet_parser.xtce.definitions instead (nested under xtce now).")
+
+
+class XtcePacketDefinition(definitions.XtcePacketDefinition):
+    def __init__(self, xtce_document, **kwargs):
+        warnings.warn("This class is deprecated. To load a packet definition from a file "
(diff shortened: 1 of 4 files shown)


space_packet_parser/definitions.py
```diff
@@ -0,0 +1,14 @@
+# NOTE: This entire file is deprecated and here just to maintain backwards compatibility.
+# The functionality has been moved to space_packet_parser.xtce.definitions.XtcePacketDefinition
+import warnings
+
+from space_packet_parser.xtce import definitions
+
+
+class XtcePacketDefinition(definitions.XtcePacketDefinition):
+    def __init__(self, xtce_document, **kwargs):
```
greglucas, line 9: previously there was an ns keyword, that I wasn't entirely clear how to transition over to the new style. I don't use the ns mapping in any of my projects, so this at least helps me for now. But happy to add that capability in if someone can tell me what I need to do.

https://github.com/lasp/space_packet_parser/blob/66aef0906ab209aabf0bb97636ddc66e449b6a70/space_packet_parser/definitions.py#L66C13-L66C38

medley56 (approved):
Nice. This is good user support.

space_packet_parser/definitions.py
medley56: The ns keyword was a way to pass in a namespace mapping dict but now that gets parsed automatically out of the XML document if the root element contains any `xmlns=<uri>` or `xmlns:prefix=<uri>` attributes. We're leaning more heavily on users to provide valid XML documents and manage their own namespace mappings, which goes hand in hand with #129

PR #148, opened 2025-02-21
Improve error message for float returned from LinearAdjuster

medley56:
LinearAdjusters are only ever intended to scale integers. This error is raised if that value is not an int for some reason. It's unlikely but this error message is preferable to just forcing integer math, which may create very confusing parsing bugs.

### DIFF
--- a/space_packet_parser/xtce/encodings.py
+++ b/space_packet_parser/xtce/encodings.py
@@ -75,30 +75,28 @@ class DataEncoding(common.AttrComparable, common.XmlObject, metaclass=ABCMeta):
             Function object that adjusts a SizeInBits value by a linear function or None if no adjuster present
         """
         if (linear_adjustment_element := parent_element.find('LinearAdjustment')) is not None:
-            slope = (int(linear_adjustment_element.attrib['slope'])
-                     if 'slope' in linear_adjustment_element.attrib else 0)
-            intercept = (int(linear_adjustment_element.attrib['intercept'])
-                         if 'intercept' in linear_adjustment_element.attrib else 0)
+            # The XTCE XSD defines slope and intercept as "doubles" so we treat them as floats
+            # Often, the result of this adjustment is assumed to be an integer number of bits, for adjusting a size
+            # (e.g. from bytes to bits) but it's not necessarily the case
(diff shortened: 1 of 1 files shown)


greglucas (commented):
I'm not sure this is correct that it needs to be an integer... I think this is where it depends on whether it is applied to an `IntegerParameter` or `FloatParameter`.
From the spec, these should be floats above this section of code.
https://www.omg.org/spec/XTCE/20180204/SpaceSystem.xsd

```
<complexType name="LinearAdjustmentType">
<annotation>
<documentation xml:lang="en">A slope and intercept may be applied to scale or shift the value of the parameter in the dynamic value</documentation>
</annotation>
<attribute name="slope" type="double"/>
<attribute name="intercept" type="double" default="0"/>
</complexType>
```

greglucas:
> My only hesitation is putting a check like this in the hot loop.

I definitely agree with not putting these checks into the processing loop. I think at that point we just throw our hands up and say you are the one who got yourself into that mess!

space_packet_parser/xtce/encodings.py
```diff
@@ -75,30 +75,28 @@ def _get_linear_adjuster(parent_element: ElementTree.Element) -> Union[callable,
...
-                     if 'slope' in linear_adjustment_element.attrib else 0)
-            intercept = (int(linear_adjustment_element.attrib['intercept'])
-                         if 'intercept' in linear_adjustment_element.attrib else 0)
+            # The XTCE XSD defines slope and intercept as "doubles" so we treat them as floats
+            # Often, the result of this adjustment is assumed to be an integer number of bits, for adjusting a size
+            # (e.g. from bytes to bits) but it's not necessarily the case
+            slope = (float(linear_adjustment_element.attrib['slope'])
+                     if 'slope' in linear_adjustment_element.attrib else 0.0)
+            intercept = (float(linear_adjustment_element.attrib['intercept'])
+                         if 'intercept' in linear_adjustment_element.attrib else 0.0)
```
greglucas, line 84:
```suggestion
            slope = float(linear_adjustment_element.attrib.get('slope', 0))
            intercept = float(linear_adjustment_element.attrib.get('intercept', 0))
```

greglucas (approved):

PR #149, opened 2025-02-21
CI: Try codecov with OIDC and avoid tokens

greglucas:
Trying this again now that we have successful codecov on main to see if that helps.


medley56:
Oh, nice! It looks like it worked! Am I missing anything?

greglucas:
Yeah, looks good to me too :+1:

PR #150, opened 2025-02-21
PERF: Use direct bitshift logic instead of _extract_bits function

greglucas:
int.from_bytes() is about 2x slower than a direct bitshift of two values, and we avoid some other checks as well.

```
python -m timeit "b=b'\x01\x02\x03\x04\x05\x06'*100; int.from_bytes(b[-2:], 'big')" 
5000000 loops, best of 5: 80.2 nsec per loop

python -m timeit "b=b'\x01\x02\x03\x04\x05\x06'*100; (b[-2] << 8) | b[-1]"
5000000 loops, best of 5: 47.8 nsec per loop
```

### DIFF
--- a/space_packet_parser/packets.py
+++ b/space_packet_parser/packets.py
@@ -97,10 +97,7 @@ class RawPacketData(bytes):
         Section 4.1.3.5.3 The length count C shall be expressed as:
         C = (Total Number of Octets in the Packet Data Field) – 1
         """
-        # This has already been parsed previously to give us the length of the packet
-        # so avoid the extract_bits call again and calculate it based on the length of the data
-        # Subtract 6 bytes for the header and 1 for the length count
-        return len(self) - RawPacketData.HEADER_LENGTH_BYTES - 1
+        return (self[4] << 8) | self[5] + 1
 
     @cached_property
     def header_values(self) -> tuple[int, ...]:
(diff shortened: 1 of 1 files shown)


space_packet_parser/packets.py
```diff
@@ -97,10 +97,7 @@ def data_length(self) -> int:
         Section 4.1.3.5.3 The length count C shall be expressed as:
         C = (Total Number of Octets in the Packet Data Field) – 1
         """
-        # This has already been parsed previously to give us the length of the packet
-        # so avoid the extract_bits call again and calculate it based on the length of the data
-        # Subtract 6 bytes for the header and 1 for the length count
-        return len(self) - RawPacketData.HEADER_LENGTH_BYTES - 1
+        return (self[4] << 8) | self[5] + 1
```
greglucas, line 100: I wasn't sure which value we want to return here. The "C" (value inside the packet) or "C + 1" (the true length of the data field). I can see arguments for either of these values.

medley56: I think we want the CCSDS defined value, C because that's the value encoded in the field that this is accessing, following the pattern of the other property accessors. If someone wants the total length of the packet, that's `len(packet_data)`. If they want the total length minus the header, that's `len(packet_data) - RawPacketData.HEADER_LENGTH_BYTES` or `packet_data.data_length + 1`

greglucas, line 100:
```suggestion
        return (self[4] << 8) | self[5]
```

medley56 (approved):

PR #153, opened 2025-02-27
Package for Anaconda

medley56:
This PR includes a few different improvements. The core change is packaging for Conda and an actions workflow for distributing the artifacts.

Additionally:
- Poetry's pyproject.toml spec has changed to be more aligned with PEP, so that's been updated
- MAYBE?: add a tool to ensure the metadata is consistent between pyproject.toml, meta.yaml, and CITATION.cff. TBD if I actually add this.

## Checklist


pyproject.toml
```diff
@@ -94,9 +123,23 @@ filterwarnings = [
     "ignore:You are encoding a BooleanParameterType:UserWarning"
 ]
 
+# Note: Ruff uses the "nearest" configuration file at runtime so configurations may be overridden
+# e.g. see ruff.toml in the tests directory
```
greglucas, line 127: The ruff.toml looks like basically the same setup except for the "S" selection. What about adding a `per-file-ignores` clause here instead of a separate `ruff.toml` to try and keep everything in one place configuration-wise?

medley56: Oooh nice. I didn't know about that option. Way nicer.

medley56:
I think check-manifest looks neat but it's not what I'm looking for. I want to make sure all the metadata matches between CITATION.cff, pyproject.toml, and the new conda recipe, meta.yaml.

medley56:
I've tested the release process using a `test-release/X.Y.Z` style tag, which makes our release pipeline push to Test PyPI and now also push to Anaconda with a `test-release` tag on the distribution. You can install the package with
```
conda install -c lasp/label/test-release space_packet_parser
```

.github/workflows/release.yml
```diff
@@ -100,12 +101,138 @@ jobs:
...
+          spp --version
+          conda deactivate
+
+      - name: Upload Conda Build Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: conda-package
+          path: ${{ env.CONDA_BLD_PATH }}/**/space_packet_parser-*.conda
+
+  # Job that uploads conda package to Anaconda.org on the LASP channel
```
greglucas, line 151: These two jobs for the official / test channels are nearly identical. I'd suggest merging them together except for the difference. A suggestion would be to add the if statement to the final step and just have two different steps one for each different upload path.

check_metadata.py
greglucas, line 1: Put this under a subdirectory rather than at the root of the project? Possibly under `ci/` or `tools/` for helper scripts.

greglucas (commented):
Looks good to me. Just a few minor suggestions you can take/leave.

check_metadata.py
medley56: Good idea. This also reminded me that the filepathing is fragile in this script because it is based on the working directory of whoever is running it. I'll put it somewhere specific and make the paths relative to the script location.

.github/workflows/release.yml
medley56: Good call. I was patterning this from the PyPI upload jobs, which require running in different environments for OIDC with PyPI. Anaconda doesn't have that restriction though.

PR #160, opened 2025-03-08
Add spp.create_packet_list function

medley56:
## Checklist

### DIFF
--- a/space_packet_parser/__init__.py
+++ b/space_packet_parser/__init__.py
@@ -1,4 +1,5 @@
 """Space Packet Parser"""
+from collections.abc import Iterable
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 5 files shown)


space_packet_parser/__init__.py
```diff
@@ -15,6 +16,43 @@ def load_xtce(filename: Union[str, Path]) -> XtcePacketDefinition:
...
+    """Directly create a list of Packet objects directly from one or more binary files and an XTCE definition
+
+    Parameters
+    ----------
+    packet_files : Union[str, Path, Iterable[Union[str, Path]]]
+        Packet files
+    xtce_packet_definition : Union[str, Path, xtce.definitions.XtcePacketDefinition]
+        Packet definition for parsing the packet data
+    packet_generator_kwargs : Optional[dict]
+        Keyword arguments passed to `XtcePacketDefinition.packet_generator()`
```
greglucas, line 38: Do we want to add this option if we are thinking about getting rid of the generator on the definition? I'm wondering if we want to accept a generator instead here, or call this method `create_ccsds_packet_list` or something like that?

I'm still torn on how this should all behave and how to make it simple and general at the same time.

medley56: If we accept a custom generator function then I think we want to allow the user to pass kwargs to it. We certainly need it for the `ccsds_generator`, right? One option could be to disallow additional kwarg passing and tell the user they have to pass the generator as a `functools.partial` with options already set. I'm not sure what is easier for people.

greglucas: Thinking broader about what it means to be generic to any packet type with XTCE parsing. We probably shouldn't provide a generator like I mentioned, instead the XTCE parsing would yield the packets that it produces during its parsing. This would also mean removing the ccsds_generator from the `definition.packet_generator()` routine.

medley56: Linking to #140.
I'm still not understanding how we would decide how many bytes to read from the source before we instantiate a `Packet` for parsing without letting users define that logic for themselves. For CCSDS it's obvious but it's not obvious to me how we would do it without a CCSDS header to tell us how many bytes are in the packet. e.g. How many bytes should go in the `binary_data` attribute of a new `Packet` object?

PR #161, opened 2025-03-08
Improve unit test coverage

medley56:
I noticed that we had low-ish coverage on a couple modules, mostly on the error-handling since most of the tests are run correctly. I found some locations where we were checking for exception cases in `evaluate` methods when we should have been pre-checking in either init or `from_xml`. 

I added some additional coverage to the packets.py module but I'm thinking that will probably need rework after @greglucas 's refactoring in #140 

## Checklist
- [NA] The changelog.md has been updated

### DIFF
--- a/space_packet_parser/xtce/comparisons.py
+++ b/space_packet_parser/xtce/comparisons.py
@@ -1,5 +1,4 @@
 """Matching logical objects"""
-import warnings
 from abc import ABCMeta, abstractmethod
 from collections import namedtuple
 from typing import Any, Optional, Union
@@ -55,8 +54,13 @@ class MatchCriteria(common.AttrComparable, common.XmlObject, metaclass=ABCMeta):
 class Comparison(MatchCriteria):
     """<xtce:Comparison>"""
 
-    def __init__(self, required_value: str, referenced_parameter: str,
-                 operator: str = "==", use_calibrated_value: bool = True):
+    def __init__(
+            self,
+            required_value: str,
+            referenced_parameter: str,
+            operator: str = "==",
+            use_calibrated_value: bool = True
+    ):
         """Constructor
 
         Parameters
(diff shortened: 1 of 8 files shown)


space_packet_parser/xtce/comparisons.py
```diff
@@ -389,21 +383,8 @@ def evaluate(self,
...
-                return packet[parameter_name] if use_calibrated \
-                    else packet[parameter_name].raw_value
-            except KeyError as e:
-                raise ComparisonError(f"Attempting to perform a Condition evaluation on {self.left_param} but "
-                                      "the referenced parameter does not appear in the hitherto parsed data passed to "
-                                      "the evaluate method. If you intended a comparison against the raw value of the "
-                                      "parameter currently being parsed, unfortunately that is not currently supported."
-                                      ) from e
-
-        # TODO: Consider allowing one of the parameters to be the parameter currently being evaluated.
```
greglucas, line 402: I thought this did work since I'm using it in other places already? Maybe that is only in a different "Comparison"... Or is the "parameter_name" the same as `self`'s name in this case so it works because of that? Ultimately I'm wondering if we can get rid of the `current_parsed_value` since it is ignored in several places and somehow rely on the comparison to the value in the packet.

tests/unit/test_packets.py
```diff
@@ -204,3 +206,36 @@ def test__extract_bits(start, nbits):
...
+
+    # From socket
+    send, recv = socket.socketpair()
+    send.send(test_packet)
+    assert next(packets.ccsds_generator(recv))
+    send.close()
+    recv.close()
+
+    # From bytes
+    # This covers show_progress conditional code and also the end of the iterator
```
greglucas, line 228: I'm curious whether we actually want the `show_progress` flag? Could we point users to `tqdm` or another progressbar package instead?

greglucas (approved):
Yeah, one of us will have to fix some merge conflicts, but overall this doesn't seem like it will cause too much headache `CCSDSPacket` -> `Packet` mostly.

tests/unit/test_packets.py
medley56: Sure! Let's investigate this in #162

space_packet_parser/xtce/comparisons.py
medley56: I'm going to add a TODO item in place of this removed error handling with a reference to a new ticket.

medley56: See #163

PR #173, opened 2025-08-02
MNT: Add better error message for XTCE from file-like

### LINKED ISSUE #172: Add XtceDefinition(str/path) warning
In the old `XtceDefinition` location we point people to the new methods, but I still ended up making the mistake in my port of doing
`definition.XtceDefinition(str_path_to_file)`, which then had an error raised later when iterating through the container names. We can intercept this early and give a nicer warning to users of which method to use to instantiate from a file.

greglucas:
closes #172

## Checklist


medley56 (approved):
Nice QOL improvement for people migrating from old versions. Merging.

PR #177, opened 2025-08-05
Allow numpy>=1.26

### LINKED ISSUE #171: Allow numpy <2.0.0
Lots of libraries are still using numpy v1. It's pretty strict to require numpy >2.0.0 with our xarray dependency. Make sure everything still works though.

medley56:
There are a lot of libraries that have pinned numpy<2.0.0. Just allowing the last v1 of numpy improves our compatibility greatly. Unfortunately going back beyond 1.26, we run into build system incompatibilities, I think because numpy<=1.25 relies on setuptools and setuptools has seen significant changes in the last few years but there's no documentation on what version is required because when 1.25 was developed, setuptools had not yet become incompatible and setuptools releases major versions every week it seems.

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -25,7 +25,7 @@ requirements:
     - click>=8.0
     - rich>=13.0
     - xarray>2024.0.0
-    - numpy>=2.0.0
+    - numpy>=1.26.0
 
 test:
   imports:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,7 @@ spp = "space_packet_parser.cli:spp"
 [project.optional-dependencies]
 xarray = [
     "xarray>2024.0.0",
-    "numpy>=2.0.0"
+    "numpy>=1.26.0"
 ]
 examples = [
     "matplotlib>=3.4"


greglucas (approved):
> There are a lot of libraries that have pinned numpy<2.0.0

Really? I think almost all major libraries have been updated at this point.

> Unfortunately going back beyond 1.26, we run into build system incompatibilities

You aren't or shouldn't be building numpy though, you are just using the numpy API. Were you testing this with a newer Python version? Once you start going back in numpy versions then you also have to test with older Python versions, so it should be a min-everything (min-python, min-numpy, min-xarray, etc.) build and not a Python 3.13 + numpy 1.x if that makes sense.

greglucas:
> Brandons Curryer library requires numpy <2.0.0 (I created an issue there).

xref? https://github.com/lasp/curryer/issues

> I didn't look too deep into it. But we don't install setuptools at all and it seems like numpy 1.25 requires something out of setup tools or pkgconfig or some other build library to work. 1.26 is when they started using meson I think.

My point is that this package shouldn't need to worry about installing the dependency. Linux distributors, conda distributors, or wheels can handle that. Once someone figures out how to install it, then does this package work with that version.

The following works fine for me locally.

```bash
python3.10 -m venv .env-310
source .env-310/bin/activate
pip install . "numpy==1.25"
```

medley56 (commented):
I stand corrected. *Now* I have created a Curryer issue: https://github.com/lasp/curryer/issues/20

Apparently this is a Python 3.12 issue with the deprecation of `pkgutil.ImpImporter` breaking the numpy install process for numpy<1.26. I was also missing setuptools from my venv. Once I installed setuptools I got the error in this Numpy issue: https://github.com/numpy/numpy/issues/23808#issuecomment-1722440746

PR #180, opened 2025-08-12
129 add a schema validator checker

medley56:
## Summary

  This PR implements comprehensive XTCE document validation capabilities for Space
  Packet Parser, adding both XSD schema validation and structural validation to help
  ensure XTCE documents are correct and will work properly for parsing packets.

  ## Changes Made

  ### Core Validation Implementation
  - **Added `space_packet_parser/xtce/validation.py`** (441 lines) - Complete
  validation system with:
    - `ValidationLevel` enum (schema, structure, all)
    - `ValidationError` and `ValidationResult` dataclasses for structured error
  reporting
    - `validate_xtce()` function as main entry point
    - Schema validation against XSD with automatic schema downloading
    - Structural validation for XTCE-specific business rules
    - Reference integrity checking (Parameters, ParameterTypes, Containers)

  ### API Integration
  - **Updated `space_packet_parser/__init__.py`** - Exported `validate_xtce` function
   for public API
  - **Updated `space_packet_parser/xtce/__init__.py`** - Renamed constants:
    - `DEFAULT_XTCE_NS_PREFIX` → `STANDARD_XTCE_NS_PREFIX`
    - `DEFAULT_XTCE_NSMAP` → `STANDARD_XTCE_NSMAP`

  ### CLI Enhancement
  - **Updated `space_packet_parser/cli.py`** - Added `spp validate` command with:
    - File path argument for XTCE documents
    - `--level` option (schema/structure/all)
    - `--timeout` option for schema downloads
    - Rich console output with color-coded results
    - Exit codes for CI/CD integration

  ### Improved Namespace Handling
  - **Enhanced `space_packet_parser/xtce/definitions.py`** - Major improvements to 
  namespace awareness:
    - Automatic XTCE namespace detection from XML documents
    - Better handling of documents without namespaces (with warnings)
    - Improved error messages for multiple XTCE namespaces
    - Fixed namespace URI vs prefix handling
    - Added validation warnings for missing XTCE namespaces

  ### Configuration Updates
  - **Updated `pyproject.toml`** - Added Ruff ignore rule for f-strings in docstrings
  - **Updated `docs/source/users.md`** - Added comprehensive validation documentation
   with examples

  ### Test Infrastructure
  - **Added `tests/unit/test_xtce/test_validation.py`** (348 lines) - Comprehensive 
  test suite:
    - Schema validation tests (valid/invalid documents)
    - Structural validation tests (reference integrity)
    - Error handling and edge cases
    - Timeout and network error scenarios
  - **Enhanced `tests/unit/test_xtce/test_definitions.py`** - Added 
  validation-related tests
  - **Updated `tests/integration/test_cli.py`** - Added CLI validation command tests
  - **Updated `tests/conftest.py`** - Updated constant references

  ### Test Data Improvements
  - **Fixed XML structure** in multiple test files to be valid XTCE:
    - `tests/test_data/ctim/ctim_xtce_v1.xml` - Fixed namespace and structure issues
    - `tests/test_data/test_xtce_no_namespace.xml` - Removed namespace for testing
    - Multiple other test files updated with proper namespace declarations

  ## Key Features

  1. **Schema Validation**: Validates against official XTCE XSD schemas with 
  automatic downloading
  2. **Structural Validation**: Checks XTCE-specific rules like reference integrity 
  and unused elements
  3. **Flexible API**: Support for file paths, Path objects, or ElementTree objects
  4. **CLI Integration**: New `spp validate` command for easy validation from command
   line
  5. **Robust Error Reporting**: Structured error objects with line numbers and XPath
   locations
  6. **Namespace Awareness**: Improved handling of XML namespaces with automatic 
  detection

  ## Test Coverage

  - 23+ new test cases covering all validation scenarios
  - Integration tests for CLI command
  - Error handling and edge case coverage
  - Network timeout and error scenarios

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,20 @@
+# Codecov configuration for space_packet_parser
+# https://docs.codecov.com/docs/codecovyml-reference
+
+coverage:
+  status:
+    project:
+      default:
+        target: 95%  # Target coverage percentage
+        threshold: 5%  # Allow 5% drop without failing
+        if_no_uploads: error  # Fail if no coverage data
+        only_pulls: false  # Check coverage on all commits
+    patch:
+      default:
+        target: 85%  # Lower threshold for new code
+        threshold: 5%
+        only_pulls: true  # Only check patch coverage on PRs
+
+  ignore:
+    - "scripts/**/*"  # Don't include utility scripts
+    - "docs/**/*"  # Don't include documentation
--- a/space_packet_parser/xtce/__init__.py
+++ b/space_packet_parser/xtce/__init__.py
@@ -3,7 +3,7 @@
 This module contains Python object representations of XTCE UML/XML models
 """
 
-DEFAULT_XTCE_NS_PREFIX = "xtce"  # Standard XTCE prefix using in xmlns:prefix="url" attribute
+STANDARD_XTCE_NS_PREFIX = "xtce"  # Standard XTCE prefix using in xmlns:prefix="url" attribute
 
 XTCE_1_2_XSD_URL = "https://www.omg.org/spec/XTCE/20180204/SpaceSystem.xsd"
 XTCE_1_2_XMLNS = "https://www.omg.org/spec/XTCE/20180204"
@@ -14,4 +14,4 @@ XTCE_1_1_XMLNS = "https://www.omg.org/spec/XTCE/20061101"
 # Note: There is no XSD available from omg.org for XTCE 1.0
 
 XTCE_URI = XTCE_1_2_XMLNS
-DEFAULT_XTCE_NSMAP = {DEFAULT_XTCE_NS_PREFIX: XTCE_1_2_XMLNS, "xsi": "http://www.w3.org/2001/XMLSchema-instance"}
+STANDARD_XTCE_NSMAP = {STANDARD_XTCE_NS_PREFIX: XTCE_1_2_XMLNS, "xsi": "http://www.w3.org/2001/XMLSchema-instance"}
--- a/space_packet_parser/xtce/validation.py
+++ b/space_packet_parser/xtce/validation.py
@@ -1,0 +1,595 @@
+"""XTCE document validation classes and utilities."""
+
+import hashlib
+import logging
+import os
+import platform
+import socket
+import time
+from dataclasses import dataclass, field
+from enum import Enum
+from pathlib import Path
+from typing import Any, Optional, Union
+from urllib.error import URLError
+from urllib.parse import urlparse
+from urllib.request import urlopen
+
+import lxml.etree as ElementTree
+
+logger = logging.getLogger(__name__)
+
+
+class ValidationLevel(Enum):
+    """Validation levels for XTCE documents."""
+
+    SCHEMA = "schema"  # Validated against XSD
+    STRUCTURE = "structure"  # Validated against XTCE-specific non-schema rules
+    ALL = "all"  # Both
+
+
+@dataclass
+class ValidationError:
+    """Represents a validation error or warning."""
+
+    message: str
+    error_code: str
+    xpath_location: Optional[str] = None
+    line_number: Optional[int] = None
+    column_number: Optional[int] = None
+    context: Optional[dict[str, Any]] = field(default_factory=dict)
+
+    def __str__(self) -> str:
+        """String representation of validation error."""
+        location_parts = []
+        if self.line_number is not None:
+            location_parts.append(f"line {self.line_number}")
+        if self.column_number is not None:
+            location_parts.append(f"col {self.column_number}")
+        if self.xpath_location:
+            location_parts.append(f"xpath: {self.xpath_location}")
+
+        location_str = f" ({', '.join(location_parts)})" if location_parts else ""
+        return f"{self.error_code}: {self.message}{location_str}"
+
+
+@dataclass
+class ValidationResult:
+    """Results of XTCE document validation."""
+
+    valid: bool
+    validation_level: ValidationLevel
+    errors: list[ValidationError] = field(default_factory=list)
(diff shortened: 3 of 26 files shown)


space_packet_parser/xtce/validation.py
```diff
@@ -0,0 +1,441 @@
...
+        )
+        schema_content = re.sub(
+            r'\s*<attribute\s+ref="xml:base"></attribute>\s*',
+            "\n\t\t\t\t<!-- xml:base attribute removed for lxml compatibility -->\n\t\t\t\t",
+            schema_content,
+        )
+    return schema_content
+
+
+def _load_schema(schema_url: str, timeout: int = 30) -> tuple[ElementTree.XMLSchema, str]:
```
greglucas, line 123: Is this something worth shipping with the package itself? Download it and redistribute it so you can reference it locally rather than having to go retrieve it.

space_packet_parser/xtce/validation.py
```diff
@@ -0,0 +1,441 @@
...
+    ----------
+    xml_tree: ElementTree.ElementTree
+        Parsed XML tree of the XTCE document
+
+    Returns
+    -------
+    ValidationResult
+        Structural validation results
+    """
+    start_time = time.time()
```
greglucas, line 276:
```suggestion
    start_time = time.perf_counter()
```

space_packet_parser/xtce/validation.py
```diff
@@ -0,0 +1,441 @@
...
+    timeout : int
+        Timeout in seconds for schema downloads
+
+    Returns
+    -------
+    ValidationResult
+        Validation results with errors, warnings, and metadata
+    """
+    try:
+        validation_level = ValidationLevel(level.lower())
```
greglucas, line 397: This seems like you should just raise immediately and not do this add_error business. If someone isn't using the function with the right calling arguments, they won't want to go and need to look into the error attributes to see something errored.

space_packet_parser/xtce/validation.py
```diff
@@ -0,0 +1,441 @@
...
+        result.add_error(f"Error during structural validation: {e}", "STRUCTURAL_VALIDATION_ERROR")
+    finally:
+        result.validation_time_ms = (time.time() - start_time) * 1000
+
+    return result
+
+
+def validate_xtce(
+    xml_source: Union[str, Path, ElementTree.ElementTree],
+    level: str = "schema",
```
greglucas, line 374: Default to "all"?

greglucas (commented):
High-level this looks OK. Some thoughts here:

If you are calling "validate_xtce()", do I really want to have to check the return after it? I'm more used to it raising if it is not valid, and then continuing on if it is. So by calling validate I kind of want you to raise an `InvalidXTCEError`. That could then contain all of the helpful information and maybe even print out within the message etc. That is a pretty large structural change from this approach, so figured I would put that at the top-level comment.

If you do keep it this current way with a "result" return, would you want to add an ability to make it evaluate as a bool so someone could say if `if not validate_xtce()` and go into the block that way?

Is there a way to automatically fix any of these suggestions and allow someone to do something on the command line like: `--validate --fix` which would overwrite or write a new version of the XTCE with the suggested changes/updates? That should be a future ticket, just a thought I had while reviewing this.

codecov.yml
```diff
@@ -0,0 +1,24 @@
...
+        if_no_uploads: error  # Fail if no coverage data
+        only_pulls: false  # Check coverage on all commits
+    patch:
+      default:
+        target: 85%  # Lower threshold for new code
+        threshold: 5%
+        only_pulls: true  # Only check patch coverage on PRs
+
+  ignore:
+    - "tests/**/*"  # Don't include test files in coverage
```
greglucas, line 19: I have heard it argued that this is actually the most important coverage of all :) If your tests aren't actually running, then you are in trouble! (If you write the same test_my_feature() function in the same file on accident one will shadow the other and not actually get run.

space_packet_parser/xtce/__init__.py
```diff
@@ -3,7 +3,7 @@
 This module contains Python object representations of XTCE UML/XML models
 """
 
-DEFAULT_XTCE_NS_PREFIX = "xtce"  # Standard XTCE prefix using in xmlns:prefix="url" attribute
+STANDARD_XTCE_NS_PREFIX = "xtce"  # Standard XTCE prefix using in xmlns:prefix="url" attribute
 
 XTCE_1_2_XSD_URL = "https://www.omg.org/spec/XTCE/20180204/SpaceSystem.xsd"
 XTCE_1_2_XMLNS = "https://www.omg.org/spec/XTCE/20180204"
```
greglucas, line 9: Add the new 1_3 version too?

greglucas (approved):
It is getting quite big for me to review now, but overall I think this is OK since it is only validation / one-time-use kinds of things, so I'll put down a few notes after trying it out.

* I'm not sure how much the validate() function will actually be used in practice, but I would have expected `raise_on_error=True` to be the default. i.e. I can just see someone doing `spp.validate_xtce(my_file) ... carry on because it is valid`
* If you are including the SpaceSystem.xsd file in the test_data, I wonder if it'd be worthwhile just including it with the actual release too at that point. It is rarely changed and ~300KB, so not the end of the world. There is a lot of caching/downloading logic involved. Doesn't make a huge difference to me either way though.
* I am getting errors saying the xsi header needs to be specified to validate an XTCE document. Is there a way you could validate a given version provided by the user or something like that instead of requiring that to be in the file itself? I realize maybe that isn't valid spec, but seems odd to fail with an error in that case to me too because our files are currently working just fine.
* I like the cli coloring and formating :)

codecov.yml
medley56: Good point!

PR #181, opened 2025-08-13
Add ruff format to pre-commit-config.yaml

medley56:
Just formatting changes


greglucas (approved):
Another nice thing to do would be to add a git-blame-ignore-revs file so we can go back through blame and avoid this formatting commit. I think that might have to be a separate PR though.

https://github.com/SciTools/cartopy/blob/main/.git-blame-ignore-revs

greglucas:
> I don't think we need this because pre-commit should prevent you from ever committing bad formatting, right?

This helps with viewing git blame, so you don't see this as a "changed" file when going through history.

PR #217, opened 2025-11-06
FIX: CCSDS packet generator bit logic needs concatentation before addition

greglucas:
There was an error in the bitshifting logic where there was a 1 added to the right-most byte before or'ing the two bytes together. This causes an error only on packets with data size greater than 2**8 and near a power of 2 boundary.

@medley56 I'd like to get this in a release ASAP because it is a pretty bad bug we are encountering when I am trying to update to the latest version (Doh! My oversight here forgetting parens). Do we want to backport this to a release bugfix branch to do a v6.0.1 release?

## Checklist

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -5,6 +5,11 @@ list and release milestones.
 ## Version Release Notes
 Release notes for the `space_packet_parser` library
 
+### v6.0.1 (unreleased)
+
+- BUGFIX: Incorrect bitshift logic for ccsds generator packet length creation
+  in very specific circumstances (only if large data on power of 2 boundary)
+
 ### v6.0.0 (released)
 - *BREAKING*: `XtcePacketDefinition` no longer accepts a file object as input.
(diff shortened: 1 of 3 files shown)


medley56 (changes requested):
Let's get this into a patch/bugfix branch based on the commit tagged as 6.0.0. We'll put the patch version 6.0.1 tag on that commit and then merge that branch back into main so all the changes are in main for 6.1.0. 

Does that make sense? This way we can get 6.0.1 released today with just the bug fix included.

medley56 (approved):
We can merge this and handle the metadata version updates in a subsequent PR or you can update them here. The version in pyproject.toml and meta.yaml just needs to be updated before we tag the commit we want to release.

greglucas:
I think we should keep them separate so that they are different commits. They really have different purposes, and this is the commit we want to cherry-pick around if anything and we don't want to carry around the other version updates with it.

greglucas:
I just don't want to merge the commits together. I think they should be two separate commits, they serve two different purposes. I'll make another PR with those updates to this branch.

PR #258, opened 2026-05-28
Issues/47 support xtce unitset

### LINKED ISSUE #47: Support xtce:UnitSet
XTCE allows for compound units by specifying multiple `<xtce:Unit>` elements for a ParameterType. This is discussed in section 4.3.2.2.4 of the XTCE element description green book, here: https://public.ccsds.org/Pubs/660x1g2.pdf (p. 65).

A compound unit will have to be represented in a ParsedDataItem. This could be solved by making the `unit` attribute of `ParsedDataItem` either a string or an iterable of strings (e.g. a tuple or list) to support compound units.

blakedehaas:
## Summary

Adds support for multiple `<xtce:Unit>` elements within a `<xtce:UnitSet>`.

Previously, only a single unit was supported. This change allows compound units to be
parsed and serialized in accordance with the XTCE specification.

## Changes

- Updated `ParameterType.get_units` to return:
  - `str` for a single unit
  - `tuple[str, ...]` for multiple units
- Updated `to_xml` to correctly serialize multiple `<Unit>` elements
- Updated type hints and docstrings for consistency

## Testing

- Added test cases for multiple `<xtce:Unit>` support in:
  - `IntegerParameterType`
  - `FloatParameterType`
  - `BinaryParameterType`
- Verified round-trip XML parsing and serialization
- All tests pass (448 total), no regressions observed

### DIFF
--- a/space_packet_parser/xtce/parameter_types.py
+++ b/space_packet_parser/xtce/parameter_types.py
@@ -16,7 +16,7 @@ from space_packet_parser.xtce import calibrators, encodings
 class ParameterType(common.AttrComparable, common.XmlObject, metaclass=ABCMeta):
     """Abstract base class for XTCE parameter types"""
 
-    def __init__(self, name: str, encoding: encodings.DataEncoding, unit: str | None = None):
+    def __init__(self, name: str, encoding: encodings.DataEncoding, unit: str | tuple[str, ...] | None = None):
         """Constructor
 
         Parameters
@@ -25,8 +25,8 @@ class ParameterType(common.AttrComparable, common.XmlObject, metaclass=ABCMeta):
             Parameter type name. Usually something like 'MSN__PARAM_Type'
         encoding : DataEncoding
             How the data is encoded. e.g. IntegerDataEncoding, StringDataEncoding, etc.
-        unit : Optional[str]
-            String describing the unit for the stored value.
+        unit : Optional[Union[str, tuple[str, ...]]]
+            Unit for stored value. String for a single unit or tuple of strings for compound units.
         """
         if name is None:
             raise ValueError("Parameter Type name attribute is required.")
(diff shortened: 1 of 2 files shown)


space_packet_parser/xtce/parameter_types.py
```diff
@@ -282,10 +291,15 @@ def to_xml(self, *, elmaker: ElementMaker) -> ElementTree.Element:
...
+
         param_type_element = getattr(elmaker, self.__class__.__name__)(name=self.name)
 
         if self.unit:
-            param_type_element.append(elmaker.UnitSet(elmaker.Unit(self.unit)))
+            if isinstance(self.unit, tuple):
+                unit_elements = [elmaker.Unit(u) for u in self.unit]
+                param_type_element.append(elmaker.UnitSet(*unit_elements))
+            else:
+                param_type_element.append(elmaker.UnitSet(elmaker.Unit(self.unit)))
```
medley56, line 303: Add a unit test that covers these lines. This method is used when we serialize XML so it should live along side the other tests for that functionality.

medley56 (changes requested):
Nice. One test coverage request.

space_packet_parser/xtce/parameter_types.py
blakedehaas: Complete 
code_coverage

tests/unit/test_xtce/test_parameter_types.py
```diff
@@ -517,6 +549,26 @@ def test_enumerated_parameter_type(elmaker, xtce_parser, xml_string: str, expect
...
+        </xtce:SizeInBits>
+    </xtce:BinaryDataEncoding>
+</xtce:BinaryParameterType>
+""",
+            parameter_types.BinaryParameterType(
+                name="TEST_PARAM_Type",
+                unit=("m", "s"),
+                encoding=encodings.BinaryDataEncoding(fixed_size_in_bits=128),
+            ),
+        ),
```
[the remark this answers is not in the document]
medley56: Yep! Good catch. @blakedehaas Let's get the round tripping of the unit set tested. Both parsing from XML and serialization to XML. You addressed the parsing side above but we should also have a serialization test to prove that we serialize multiple unit elements correctly.

medley56 (approved):

PR #267, opened 2026-07-20
Harden XTCE schema validation against local file read (CWE-73) and SSRF (CWE-918)

### LINKED ISSUE #266: path validation and ssrf
# Context

Open source contributor identified the following issue

# Driving Requirements

two issues in space_packet_parser affecting the validate_xtce() function.

# Implementation Requirements
---

**Finding 1 — SSRF via xsi:schemaLocation (CWE-918)**

validate_xtce() reads the xsi:schemaLocation attribute directly from the user-supplied XTCE document and makes an HTTP request to that URL without any validation or allowlist. An attacker can supply a malicious XTCE file to trigger outbound HTTP requests to arbitrary hosts, including cloud metadata endpoints (e.g. http://169.254.169.254/latest/meta-data/iam/security-credentials/).

Additionally, the HTTP response is cached to disk at ~/.cache/space_packet_parser/schemas/, meaning retrieved content persists after the process terminates.

Affected code: space_packet_parser/xtce/validation.py — _load_schema() and _find_schema_url()

**Finding 2 — Local File Read via absolute path in xsi:schemaLocation (CWE-73)**

If the xsi:schemaLocation attribute contains an absolute filesystem path (e.g. /etc/passwd, /home/user/.ssh/id_rsa), the code opens and reads that file without any path validation.

Affected code: space_packet_parser/xtce/validation.py — _load_schema(), line: with Path(schema_location).open('rb') as sfh

# Considerations

Finding 1 Affected code: space_packet_parser/xtce/validation.py — _load_schema() and _find_schema_url()

Finding 2 Affected code: space_packet_parser/xtce/validation.py — _load_schema(), line: with Path(schema_location).open('rb') as sfh

blakedehaas:
## Summary

Hardens XTCE schema validation against two reported vulnerabilities in `validate_xtce`. Both share one root cause: a document's `xsi:schemaLocation` is attacker-controlled but was used to read files and make network requests without restriction. Fixes #266.

## Problems being fixed

- **SSRF — CWE-918 (7.4 High).** A crafted document could point `xsi:schemaLocation` at any URL, causing the validating host to issue outbound requests — e.g. the cloud metadata endpoint `http://169.254.169.254/latest/meta-data/iam/security-credentials/` for IAM credential theft — and the response was cached to disk, persisting exfiltrated data.
- **LFI — CWE-73.** The same field could point at an absolute or traversal filesystem path (`/etc/hostname`, `../../etc/passwd`), causing arbitrary local files to be read as a "schema".

## How we fix it

The document-supplied `xsi:schemaLocation` is now treated as **untrusted**, while an operator-supplied `local_xsd` remains **trusted**. Schema resolution is:

1. **Bundled schema (offline).** The standard OMG XTCE 1.2 schema ships in the package, so the common case validates with **no network request** — this alone removes the SSRF surface for typical use.
2. **`local_xsd` (trusted).** Opened directly at any path.
3. **Allowlisted download.** Any other URL is fetched **only** if it is `https` on an allowlisted host (default `www.omg.org`). Internal/link-local targets (`169.254.169.254`, `127.0.0.1`, private ranges) are always blocked. Local filesystem paths from a document are rejected outright — use `local_xsd`.

Additional hardening:
- Downloaded content is **size-capped** and cached **only after** it validates as an XSD, so a non-schema response can never be persisted.
- New options (API + CLI, with env-var overrides): `allowed_schema_hosts` (hosts or exact URLs), `allow_insecure_http` (dangerous opt-in; host allowlist + internal-address guard still apply), `allow_schema_download`. Default download stays on but allowlisted. `DEFAULT_ALLOWED_SCHEMA_HOSTS` is exported.

## Behavior changes

- A document whose `xsi:schemaLocation` is a **local path** or a **non-allowlisted / non-https URL** is now rejected — pass `local_xsd`, extend the allowlist, or (for http) opt in explicitly.
- Absolute `local_xsd` / `--local-xsd` again works from any working directory (fixes a regression that rewrote it to a bare filename).

## Testing

New `tests/unit/test_xtce/test_validation_security.py` asserts every advisory PoC is rejected with no outbound request, plus positive coverage for the allowlist (arg/env/exact-URL), the `allow_insecure_http` opt-in still blocking internal IPs, offline bundle validation, and cache behavior. Full suite passes; docs and changelog updated.

<sub><em>PR description drafted with Claude and reviewed by Gavin (@medley56).</em></sub>

### DIFF
--- a/space_packet_parser/xtce/validation.py
+++ b/space_packet_parser/xtce/validation.py
@@ -249,8 +249,30 @@ def _load_schema(schema_location: str | Path, timeout: int = 30) -> tuple[Elemen
             logger.debug(f"Using cached schema from {cache_path}")
     # Otherwise assume a local filepath
     else:
-        with Path(schema_location).open("rb") as sfh:
-            schema_content = sfh.read()
+        schema_path = Path(schema_location)
+        
+        # Reject absolute paths (CWE-73)
+        if schema_path.is_absolute():
+            raise XtceValidationError(
+                f"Absolute filesystem paths are not allowed: {schema_location}. "
+                "Use relative paths or http/https URLs."
+            )
+        
+        # Prevent path traversal (e.g., ../../../etc/passwd)
+        try:
+            resolved = schema_path.resolve()
+            if not str(resolved).startswith(str(Path.cwd().resolve())):
+                raise XtceValidationError(
+                    f"Path traversal detected. Schema path must be within current working directory: {schema_location}"
+                )
+        except Exception as e:
+            raise XtceValidationError(f"Invalid schema path: {schema_location}") from e
+        
+        try:
+            with schema_path.open("rb") as sfh:
+                schema_content = sfh.read()
+        except FileNotFoundError as e:
+            raise XtceValidationError(f"Schema file not found: {schema_location}") from e
 
     # Fix and parse the schema content
     try:
@@ -289,6 +311,11 @@ def _find_schema_url(xml_tree: ElementTree.ElementTree) -> str:
     -------
     schema_location : str
         URL of XSD
+        
+    Raises
+    ------
+    XtceValidationError
+        If schema location is invalid or missing
     """
     # Get root element
     root = xml_tree.getroot() if hasattr(xml_tree, "getroot") else xml_tree
@@ -296,12 +323,27 @@ def _find_schema_url(xml_tree: ElementTree.ElementTree) -> str:
     # Find schema location
     try:
         schema_location_attr = root.attrib.get("{http://www.w3.org/2001/XMLSchema-instance}schemaLocation")
-        return schema_location_attr.split()[-1]
+        schema_location = schema_location_attr.split()[-1]
     except Exception:
         raise XtceValidationError(
             "No 'xsi' namespace found in document. XTCE documents must declare the 'xsi' "
             "namespace for schema validation via the 'xsi:schemaLocation' attribute."
         )
+    
+    # Reject absolute filesystem paths (CWE-73)
+    if schema_location.startswith('/'):
+        raise XtceValidationError(
+            f"Absolute filesystem paths are not allowed in xsi:schemaLocation: {schema_location}"
+        )
+    
+    # Only allow http/https URLs (CWE-918)
+    parsed = urlparse(schema_location)
+    if parsed.scheme and parsed.scheme not in ('http', 'https'):
+        raise XtceValidationError(
+            f"Only http and https URLs are allowed in xsi:schemaLocation. Got: {parsed.scheme}"
+        )
+    
+    return schema_location
 
 
 def _validate_xtce_schema(
@@ -357,7 +399,7 @@ def _validate_xtce_schema(
             for error in schema.error_log:
                 if "No matching global declaration available for the validation root." in error.message:
                     result.add_error(
-                        message="Namespace issue detected. Does the `xmlns[:xtce]=<chosen_xtce_uri>` URI on your document root element match the `targetNamespace` URI in your XSD? Typically this is http://www.omg.org/spec/XTCE/20180204.",
+                        message="Namespace issue detected. Does the `xmlns[:xtce]=<chosen_xtce_uri>` URI on your document root element match the `targetNamespace` URI in your XSD? Typically this [...]
                         error_code="INVALID_XTCE_NAMESPACE",
                         context={
                             "nsmap": xml_tree.getroot().nsmap,
--- a/space_packet_parser/xtce/validation.py
+++ b/space_packet_parser/xtce/validation.py
@@ -250,14 +250,13 @@ def _load_schema(schema_location: str | Path, timeout: int = 30) -> tuple[Elemen
     # Otherwise assume a local filepath
     else:
         schema_path = Path(schema_location)
-        
+
         # Reject absolute paths (CWE-73)
         if schema_path.is_absolute():
             raise XtceValidationError(
-                f"Absolute filesystem paths are not allowed: {schema_location}. "
-                "Use relative paths or http/https URLs."
+                f"Absolute filesystem paths are not allowed: {schema_location}. Use relative paths or http/https URLs."
             )
-        
+
         # Prevent path traversal (e.g., ../../../etc/passwd)
         try:
             resolved = schema_path.resolve()
@@ -267,7 +266,7 @@ def _load_schema(schema_location: str | Path, timeout: int = 30) -> tuple[Elemen
                 )
         except Exception as e:
             raise XtceValidationError(f"Invalid schema path: {schema_location}") from e
-        
+
         try:
             with schema_path.open("rb") as sfh:
                 schema_content = sfh.read()
@@ -311,7 +310,7 @@ def _find_schema_url(xml_tree: ElementTree.ElementTree) -> str:
     -------
     schema_location : str
         URL of XSD
-        
+
     Raises
     ------
     XtceValidationError
@@ -329,20 +328,16 @@ def _find_schema_url(xml_tree: ElementTree.ElementTree) -> str:
             "No 'xsi' namespace found in document. XTCE documents must declare the 'xsi' "
             "namespace for schema validation via the 'xsi:schemaLocation' attribute."
         )
-    
+
     # Reject absolute filesystem paths (CWE-73)
-    if schema_location.startswith('/'):
-        raise XtceValidationError(
-            f"Absolute filesystem paths are not allowed in xsi:schemaLocation: {schema_location}"
-        )
-    
+    if schema_location.startswith("/"):
+        raise XtceValidationError(f"Absolute filesystem paths are not allowed in xsi:schemaLocation: {schema_location}")
+
     # Only allow http/https URLs (CWE-918)
     parsed = urlparse(schema_location)
-    if parsed.scheme and parsed.scheme not in ('http', 'https'):
-        raise XtceValidationError(
-            f"Only http and https URLs are allowed in xsi:schemaLocation. Got: {parsed.scheme}"
-        )
-    
+    if parsed.scheme and parsed.scheme not in ("http", "https"):
+        raise XtceValidationError(f"Only http and https URLs are allowed in xsi:schemaLocation. Got: {parsed.scheme}")
+
     return schema_location
 
 
@@ -380,7 +375,7 @@ def _validate_xtce_schema(
                 except ValueError:
                     # If path is not relative to cwd, try using just the filename
                     xsd_path = Path(xsd_path.name)
-            schema_location = str(xsd_path) 
+            schema_location = str(xsd_path)
         else:
             try:
                 # Find the URL of the XSD
--- a/space_packet_parser/xtce/validation.py
+++ b/space_packet_parser/xtce/validation.py
@@ -399,7 +399,7 @@ def _validate_xtce_schema(
             for error in schema.error_log:
                 if "No matching global declaration available for the validation root." in error.message:
                     result.add_error(
-                        message="Namespace issue detected. Does the `xmlns[:xtce]=<chosen_xtce_uri>` URI on your document root element match the `targetNamespace` URI in your XSD? Typically this [...]
+                        message="Namespace issue detected. Does the `xmlns[:xtce]=<chosen_xtce_uri>` URI on your document root element match the `targetNamespace` URI in your XSD? Typically this is http://www.omg.org/spec/XTCE/20180204
                         error_code="INVALID_XTCE_NAMESPACE",
                         context={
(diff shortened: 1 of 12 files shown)


medley56 (commented):
*🤖 This review was generated with Claude and reviewed by Gavin (@medley56).*

Hi @blakedehaas — thanks for kicking this off. Getting the SSRF/LFI hardening started was the hard part, and the structure you set up (a dedicated schema-load path, the caching layer, the test scaffolding) is what we built the final version on top of. As code owner I took the branch the rest of the way, and I want to walk you through what we changed and why so nothing here is a surprise.

**The core reframe:** a document's `xsi:schemaLocation` is fully attacker-controlled whenever we validate an untrusted document, so the fix has to treat it as untrusted at every step — while keeping the operator-supplied `local_xsd` fully trusted. The original approach guarded some of the surface but left the highest-severity path open, so we re-centered everything on that trust boundary.

**What was still exploitable (and is now closed):**
- **SSRF (the 7.4 High) was not actually mitigated.** The scheme check allowed any `http`/`https` URL, so the advisory's own PoC — `http://169.254.169.254/latest/meta-data/...` — sailed through. We replaced the scheme-only check with **https-by-default + a host/URL allowlist + an internal-address guard**, so metadata/loopback/private targets are blocked outright.
- **LFI was only partly closed.** `startswith("/")` missed Windows/UNC paths, and a sibling-directory string-prefix bug let some traversals through. Document-derived local paths are now rejected wholesale (any non-`http(s)` location) — a local schema must come through `local_xsd`.

**What we added:**
- **Bundled the OMG XTCE 1.2 schema**, so the standard case validates **offline with no network request** — this removes the SSRF surface for the common path entirely and makes validation robust to OMG being unreachable.
- **Configurable, safe-by-default controls**: `allowed_schema_hosts` (hosts or exact URLs), `allow_insecure_http` (dangerous opt-in), and `allow_schema_download`, all mirrored on the CLI and with env-var overrides. Default download stays **on** but allowlisted.

**What we fixed from the first iteration:** restored absolute `local_xsd` (it was being silently rewritten to a bare filename), stopped caching un-validated bytes, added a download size cap, gave rejections accurate/surfaced error codes, and moved the `mock_schema_download` fixture into `conftest.py` (the integration test couldn't find it, so it was erroring at setup).

Full rationale is in the changelog and `docs/source/users.md`; details are in the inline comments. Happy to hop on a call if any of the decisions are worth talking through — especially the "download on-by-default but allowlisted" choice, since that was the main judgment call.

space_packet_parser/xtce/validation.py
```diff
@@ -289,25 +527,48 @@ def _find_schema_url(xml_tree: ElementTree.ElementTree) -> str:
...
+            "namespace for schema validation via the 'xsi:schemaLocation' attribute.",
+            error_code="MISSING_SCHEMA_LOCATION",
+        )
+
+    # A document-supplied location is untrusted. Anything that is not an http(s) URL is treated as
+    # a local filesystem reference and rejected to prevent local file disclosure (CWE-73). This
+    # covers absolute POSIX/Windows/UNC paths, relative paths, bare filenames, and non-http schemes
+    # such as file:// and ftp://. Host/scheme allowlist policy is enforced later, at load time,
+    # after bundled-schema resolution.
+    if urlparse(schema_location).scheme not in ("http", "https"):
```
medley56, line 555: *🤖 This comment was generated with Claude and reviewed by Gavin (@medley56).*

**LFI fix (CWE-73).** The original guard here was `schema_location.startswith("/")`, which only catches POSIX absolute paths — it missed Windows drive paths (`C:\...`), UNC paths (`\\server\share`), and left relative traversals (`../../etc/passwd`) to a downstream check that had a sibling-directory prefix bug. We inverted the logic: anything that isn't an `http(s)` URL is treated as a local reference and rejected here, up front. A local schema now has exactly one path in — the trusted `local_xsd` argument. Regression-tested against `/etc/hostname`, `../../../etc/passwd`, `C:\...`, UNC, and `file://` in `test_validation_security.py`.

space_packet_parser/xtce/validation.py
```diff
@@ -203,78 +244,275 @@ def _fix_known_schema_issues(schema_content: bytes) -> bytes:
...
+            continue
+        if "://" in entry:
+            if _normalize_url(entry) == normalized:
+                return True
+        elif entry.lower() == host:
+            return True
+    return False
+
+
+def _enforce_schema_url_policy(schema_location: str, allowed_hosts: frozenset[str], allow_insecure_http: bool) -> None:
```
medley56, line 335: *🤖 This comment was generated with Claude and reviewed by Gavin (@medley56).*

**SSRF fix (CWE-918) — the headline change.** The previous version allowed any `http`/`https` URL, which means the advisory's PoC (`http://169.254.169.254/latest/meta-data/iam/security-credentials/`) was still fully reachable — a scheme check is not an SSRF control. This helper now enforces three things together: **https only** (unless `allow_insecure_http` is explicitly set), the **host/URL allowlist** (default `www.omg.org`; entries can be bare hosts or exact URLs, matched exactly — never a suffix test), and the internal-address guard below. All three are independent, so relaxing one (e.g. enabling http for an internal mirror) doesn't reopen the others.

space_packet_parser/xtce/validation.py
```diff
@@ -203,78 +244,275 @@ def _fix_known_schema_issues(schema_content: bytes) -> bytes:
...
+    if parsed.scheme not in ("http", "https"):
+        return None
+    key = f"{parsed.netloc.lower()}{parsed.path}"
+    filename = _BUNDLED_SCHEMAS.get(key)
+    if filename is None:
+        return None
+    return _BUNDLED_SCHEMA_DIR / filename
+
+
+def _reject_internal_host(host: str | None, schema_location: str) -> None:
```
medley56, line 288: *🤖 This comment was generated with Claude and reviewed by Gavin (@medley56).*

This is the guard that deterministically kills the cloud-metadata and loopback PoCs. IP **literals** in the URL (`169.254.169.254`, `127.0.0.1`, `10.x`, etc.) are rejected as private/loopback/link-local/reserved regardless of the allowlist. We scoped it to IP literals on purpose: hostnames are governed by the allowlist, and doing DNS resolution here would add a network round-trip plus a TOCTOU/rebinding wrinkle. The residual DNS-rebinding case (an allowlisted hostname that resolves internally) is noted in the docstring as a documented, much narrower follow-up.

space_packet_parser/xtce/validation.py
```diff
@@ -203,78 +244,275 @@ def _fix_known_schema_issues(schema_content: bytes) -> bytes:
...
+        Parsed XMLSchema object and version string.
+
+    Raises
+    ------
+    XtceValidationError
+        If the schema cannot be loaded, is disallowed by policy, or cannot be parsed.
+    """
+    location = str(schema_location)
+
+    # 1. Bundled schema (offline, no policy needed — it is our own file).
```
medley56, line 476: *🤖 This comment was generated with Claude and reviewed by Gavin (@medley56).*

**Bundling (offline by default).** Resolution order is now: (1) bundled schema → (2) trusted `local_xsd` → (3) allowlisted download. Because the OMG XTCE 1.2 schema ships in the package (`xtce/schemas/`), a document referencing the standard `www.omg.org/.../SpaceSystem.xsd` URL validates from disk with **zero network calls** — verified by a test that patches `urlopen` and asserts it's never called. This both removes the SSRF surface for the common case and makes validation resilient to OMG downtime. Only XTCE 1.2 is bundled today; a 1.1 document falls through to the allowlisted-download path (OMG-hosted, so it still works). Adding 1.1 offline is a one-line `_BUNDLED_SCHEMAS` entry.

space_packet_parser/xtce/validation.py
```diff
@@ -330,25 +599,39 @@ def _validate_xtce_schema(
 
     try:
         if local_xsd:
-            # TODO: Load the XSD from local file
+            # local_xsd is trusted, caller-supplied input: open it directly, at whatever path
+            # (absolute or relative) the caller provided. No confinement is applied because this
+            # is not the attacker-controlled surface (unlike document-derived xsi:schemaLocation).
             schema_location = str(local_xsd)
```
medley56, line 605: *🤖 This comment was generated with Claude and reviewed by Gavin (@medley56).*

**Trust boundary + regression fix.** The prior version treated `local_xsd` like untrusted input: an absolute path was rewritten via `relative_to(cwd())`, falling back to just the *basename* when that failed — which silently broke absolute `--local-xsd` from any other working directory, or worse, validated against a same-named file that happened to sit in cwd. `local_xsd` is operator-supplied and trusted, so it's now opened directly at whatever path was given. The confinement effort belongs on the untrusted document path (above), not here. Restored behavior is covered by `test_validate_xtce_absolute_local_xsd_is_used_directly`.

space_packet_parser/xtce/validation.py
```diff
@@ -203,78 +244,275 @@ def _fix_known_schema_issues(schema_content: bytes) -> bytes:
...
+        ) from e
+
+    if len(schema_content) > MAX_SCHEMA_BYTES:
+        raise XtceValidationError(
+            f"Schema at {schema_location} exceeds the maximum allowed size ({MAX_SCHEMA_BYTES} bytes).",
+            error_code="SCHEMA_LOAD_ERROR",
+        )
+
+    schema, version = _parse_schema_content(schema_content, schema_location)
+    # Only cache content that validated as an XSD.
```
medley56, line 424: *🤖 This comment was generated with Claude and reviewed by Gavin (@medley56).*

**Cache hardening.** Previously the raw HTTP response was written to `~/.cache/...` *before* it was parsed, so a non-schema body (an SSRF probe response, an error page, exfiltrated data) would persist on disk — which the advisory explicitly called out. Now we download with a **size cap** (`MAX_SCHEMA_BYTES`), validate the bytes as an XSD, and only *then* write to cache. We deliberately cache the **raw** validated bytes (not the post-`_fix_known_schema_issues` version) so the download-avoidance benefit is preserved without baking a fixup into the cache — a future improvement to the fix logic still applies on the next load. Cache files are also written `0600`.

space_packet_parser/xtce/validation.py
```diff
@@ -102,9 +126,17 @@ def __str__(self) -> str:
...
+        self,
+        message: str,
+        validation_result: ValidationResult | None = None,
+        error_code: str | None = None,
+    ):
         super().__init__(message)
         self.validation_result = validation_result
+        # Optional machine-readable code so callers (e.g. _validate_xtce_schema) can map
+        # the failure onto the correct ValidationResult error_code instead of a generic one.
+        self.error_code = error_code
```
medley56, line 139: *🤖 This comment was generated with Claude and reviewed by Gavin (@medley56).*

Small but worth flagging: in the first iteration the "Path traversal detected" error was raised *inside* a `try` whose own `except Exception` immediately re-wrapped it as a generic "Invalid schema path", so the real reason never reached the caller — and every rejection from `_find_schema_url` was reported as `MISSING_SCHEMA_LOCATION` even when a location was present but disallowed. Errors now carry an explicit `error_code`, so a disallowed/blocked location surfaces as `DISALLOWED_SCHEMA_LOCATION` with its actual message, distinct from a genuinely missing one.

tests/unit/test_xtce/test_validation_security.py
```diff
@@ -0,0 +1,257 @@
...
+@pytest.mark.parametrize(
+    "location",
+    [
+        "http://169.254.169.254/latest/meta-data/iam/security-credentials/",  # cloud metadata PoC
+        "http://127.0.0.1:9998/SSRF_CONFIRMED",  # loopback PoC
+        "https://evil.example.com/schema.xsd",  # arbitrary external host
+        "ftp://example.com/schema.xsd",  # non-http scheme
+    ],
+)
+def test_document_url_ssrf_is_rejected(location):
```
medley56, line 90: *🤖 This comment was generated with Claude and reviewed by Gavin (@medley56).*

New regression suite anchored directly to the advisories: every PoC (metadata IP, loopback, arbitrary external host, absolute/relative/Windows/UNC/`file://` local paths) asserts rejection *and* that `urlopen` is never called; plus positive coverage for the allowlist (arg, env var, exact-URL entries), the `allow_insecure_http` opt-in still being blocked by the internal-IP guard, offline bundle use, and the cache behavior. Separately, the `mock_schema_download` fixture moved to `tests/conftest.py` — it lived in this module before, so `tests/integration/test_cli.py::test_validate_xtce` couldn't resolve it and was erroring at setup rather than running.

djesernik (approved):
This LGTM in terms of addressing the vulnerability.

I do think there may be some issues of semantics with the use of `host` (which can mean `example.org:8080`) where `hostname` (just `example.org`) would be more appropriate.

PR #282, opened 2026-09-13
Raise UnrecognizedPacketTypeError without assuming a PKT_APID parameter

### LINKED ISSUE #276: UnrecognizedPacketTypeError message assumes a PKT_APID parameter, raising KeyError for non-CCSDS definitions
# Context

`parse_bytes` builds its `UnrecognizedPacketTypeError` message by interpolating `packet['PKT_APID']`:

https://github.com/lasp/space_packet_parser/blob/9168a3e9b3c515cb521af22b6ddb2ece845c4a7a/space_packet_parser/xtce/definitions.py#L451-L458

That subscript assumes the packet has a parameter literally named `PKT_APID`. XTCE has no notion of the CCSDS standard, and any container in a definition may be used as the root, so this assumption does not hold in general. When it fails, the `KeyError` is raised *while constructing the exception*, so the intended `UnrecognizedPacketTypeError` is never raised at all.

Noticed while correcting the `root_container_name` docstrings in #273, which had claimed a root container "must begin with the definition of a CCSDS header in order to parse correctly". That claim is false — but this line is the one place in the parse path where it is effectively assumed to be true.

## Reproduction

A definition with no CCSDS header, an abstract root container, and an inheritor whose restriction criteria do not match:

```python
import io
import struct

from space_packet_parser import load_xtce
from space_packet_parser.exceptions import UnrecognizedPacketTypeError

XTCE = """<?xml version="1.0" encoding="UTF-8"?>
<xtce:SpaceSystem name="NoCcsds" xmlns:xtce="http://www.omg.org/spec/XTCE/20180204">
  <xtce:TelemetryMetaData>
    <xtce:ParameterTypeSet>
      <xtce:IntegerParameterType name="U8_Type" signed="false">
        <xtce:IntegerDataEncoding sizeInBits="8" encoding="unsigned"/>
      </xtce:IntegerParameterType>
    </xtce:ParameterTypeSet>
    <xtce:ParameterSet>
      <xtce:Parameter name="COUNTER" parameterTypeRef="U8_Type"/>
    </xtce:ParameterSet>
    <xtce:ContainerSet>
      <xtce:SequenceContainer name="AbstractRoot" abstract="true">
        <xtce:EntryList><xtce:ParameterRefEntry parameterRef="COUNTER"/></xtce:EntryList>
      </xtce:SequenceContainer>
      <xtce:SequenceContainer name="Child">
        <xtce:EntryList/>
        <xtce:BaseContainer containerRef="AbstractRoot">
          <xtce:RestrictionCriteria>
            <xtce:Comparison parameterRef="COUNTER" value="99" useCalibratedValue="false"/>
          </xtce:RestrictionCriteria>
        </xtce:BaseContainer>
      </xtce:SequenceContainer>
    </xtce:ContainerSet>
  </xtce:TelemetryMetaData>
</xtce:SpaceSystem>
"""

definition = load_xtce(io.BytesIO(XTCE.encode()))
data = struct.pack(">B", 7)  # COUNTER=7, so Child's criteria (== 99) fails

try:
    definition.parse_bytes(data, root_container_name="AbstractRoot")
except UnrecognizedPacketTypeError as e:
    print("intended error:", e)
```

**Expected:** `UnrecognizedPacketTypeError`, caught by the `except` clause, carrying `partial_data`.

**Actual:** the `except` clause does not fire, and the `KeyError` propagates:

```
  File "space_packet_parser/xtce/definitions.py", line 454, in parse_bytes
    f"APID={packet['PKT_APID']}.",
            ~~~~~~^^^^^^^^^^^^
KeyError: 'PKT_APID'
```

# Driving Requirements

A packet that cannot be identified should raise `UnrecognizedPacketTypeError` regardless of whether the definition happens to use CCSDS naming. Error reporting on the failure path should not itself be able to fail.

# Implementation Requirements

- `parse_bytes` raises `UnrecognizedPacketTypeError` for the abstract-container-with-no-valid-inheritors case even when the packet has no `PKT_APID` parameter.
- `partial_data` is populated in that case, since it is the main diagnostic the error exists to carry.
- The message still reports the APID when one is available, since that is genuinely useful for the common CCSDS case.
- Regression test covering a definition without a `PKT_APID` parameter.

A minimal fix is to stop subscripting on the failure path. `SpacePacket` subclasses `dict`, so `.get` is available:

```python
f"APID={packet.get('PKT_APID', 'unknown')}."
```

# Considerations

- **Severity is low but the failure mode is unpleasant.** It is invisible for ordinary CCSDS definitions, where `PKT_APID` is parsed from the header before inheritors are evaluated. It surfaces only once someone uses a non-CCSDS definition or different parameter naming — at which point the documented `try`/`except UnrecognizedPacketTypeError` pattern silently stops working and callers get a `KeyError` from inside the library.
- **`partial_data` is lost.** Because the failure happens during exception construction, the caller gets neither the intended exception type nor the partial packet, which is exactly the information needed to debug a definition that fails this way.
- **Naming, not just the CCSDS header.** Even a definition that does describe CCSDS packets hits this if it names the field something other than `PKT_APID` (`APID`, `ApID`, etc.). Nothing in XTCE mandates that spelling.
- **Worth a scan for sibling assumptions.** This is the only `PKT_APID` subscript in `definitions.py`, but other CCSDS-specific naming assumptions may exist on other error paths.
- No API change and no behavior change for existing CCSDS users — only the failure path is affected.

medley56:
Closes #276

## Problem

When an abstract container had no valid inheritors, `parse_bytes` built the `UnrecognizedPacketTypeError` message by subscripting `packet['PKT_APID']`. XTCE has no notion of the CCSDS standard, so any definition without a parameter literally named `PKT_APID` raised `KeyError` *while constructing the exception*. The documented `except UnrecognizedPacketTypeError` pattern silently stopped working and `partial_data` was lost.

## Fix

- The message now names the abstract container that had no valid inheritors.
- The APID is reported only when the packet actually has a `PKT_APID` value. If it does not, but the bytes are `CCSDSPacketBytes`, the CCSDS header printout is appended instead (same approach as the existing trailing-bits warning). Otherwise the message simply omits APID information.
- `partial_data` is populated in every case.
- `CCSDSPacketBytes.__str__` no longer raises `IndexError` for inputs shorter than the six-byte primary header. It renders the bytes it has (`CCSDSPacket Header: (incomplete, 1 of 6 bytes: 07)`), so the header printout is safe to use in diagnostic messages, including the pre-existing trailing-bits warning. Construction is unchanged.
- No API change and no behavior change for CCSDS definitions beyond the message wording.

I scanned the package for sibling CCSDS-naming assumptions on error paths. This was the only `PKT_APID` subscript; the only other CCSDS field names outside the `generators.ccsds` module are the CLI display headers, which read from `CCSDSPacketBytes` rather than parsed parameters.

## Tests

- `test_parse_bytes_unrecognized_packet_without_pkt_apid`: the reproduction from the issue. Asserts the intended exception type, that the message does not claim an APID, and that `partial_data` carries the parsed `COUNTER`.
- `test_parse_bytes_unrecognized_packet_reports_apid`: a CCSDS packet with an undefined APID still gets `APID=...` in the message and in `partial_data`.
- `test_parse_bytes_unrecognized_packet_ccsds_bytes_without_pkt_apid`: the `CCSDSPacketBytes` fallback branch, with a full header and with a single byte.
- `test_ccsds_packet_bytes_str_short_input`: `str(CCSDSPacketBytes)` on fewer than six bytes.

Full suite: 485 passed. `ruff check` and `ruff format --check` clean. Changelog entry added under Unreleased.


CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
```
medley56, line 10: ⚠️ `WARNING` — CHANGELOG.md conflicts with main after #290

#290 (8c1ae21) landed on main after this branch was cut and also added the first bullet under `[Unreleased] / ### Fixed`, so GitHub reports the PR as not mergeable. Resolution is to keep both bullets; `CITATION.cff` and `scripts/check_metadata.py` merge cleanly.

space_packet_parser/xtce/definitions.py
```diff
@@ -450,12 +450,19 @@ def parse_bytes(self, binary_data: bytes, *, root_container_name: str | None = N
...
+                        f"restriction criteria. This might mean this packet type is not accounted for in the "
+                        f"provided packet definition."
                     )
+                    # XTCE has no notion of the CCSDS standard, so a PKT_APID parameter is not guaranteed to exist.
+                    # Report the APID when it is available because it is the most useful diagnostic for CCSDS
+                    # packets, but never let the error report itself fail on a non-CCSDS definition.
+                    if "PKT_APID" in packet:
+                        message += f" APID={packet['PKT_APID']}."
+                    elif isinstance(packet.binary_data, ccsds.CCSDSPacketBytes):
+                        message += f" {packet.binary_data}."
```
medley56, line 464: 💡 `SUGGESTION` — fix the short-header IndexError in `CCSDSPacketBytes.__str__`, not here

Copilot's IndexError finding on this line reproduces: `xdef.parse_bytes(CCSDSPacketBytes(b"\x07"))` with the PR's own no-CCSDS definition raises `IndexError` from the `apid` property instead of `UnrecognizedPacketTypeError`. Two points beyond that thread:

- The pre-existing trailing-bits warning at `definitions.py:477-479` uses the same `isinstance` + `str(packet.binary_data)` pattern and has the same latent IndexError, so a guard on this line alone leaves that one in place.
- Making `CCSDSPacketBytes.__str__` (`generators/ccsds.py:42-47`) return a truncated rendering when `len(self) < HEADER_LENGTH_BYTES` fixes both call sites and keeps header-length knowledge out of `definitions.py`.

Exposure is low, since `ccsds_generator` never yields a packet shorter than six bytes, but the PR's stated goal is that the error report can never itself fail.

medley56, line 464: ✏️ `NITPICK` — CCSDS-header append is now duplicated with the trailing-bits warning

Lines 463-464 and 477-479 now contain the same `isinstance(packet.binary_data, ccsds.CCSDSPacketBytes)` + `message += f" {packet.binary_data}."` pair. If the `__str__` change is not taken, a small private helper would keep the two diagnostics from drifting.

medley56 (commented):
Fix is correct and minimal. The KeyError path from #276 is gone, partial_data is populated on every raise, and the CCSDS message now names the abstract container as well as the APID. Both new tests pass and CI is green on all 20 checks. Two things stand before merge: CHANGELOG.md conflicts with #290 which landed on main after this branch was cut, and the two open Copilot threads are both valid (I reproduced the IndexError on a short CCSDSPacketBytes; the fallback branch is Codecov's one uncovered line).

Severity tally: 0 critical, 1 warning, 1 suggestion, 1 nitpick.

space_packet_parser/xtce/definitions.py
```diff
@@ -450,12 +450,19 @@ def parse_bytes(self, binary_data: bytes, *, root_container_name: str | None = N
...
+                        f"Detected an abstract container ({current_container.name}) with no valid inheritors by "
+                        f"restriction criteria. This might mean this packet type is not accounted for in the "
+                        f"provided packet definition."
                     )
+                    # XTCE has no notion of the CCSDS standard, so a PKT_APID parameter is not guaranteed to exist.
+                    # Report the APID when it is available because it is the most useful diagnostic for CCSDS
+                    # packets, but never let the error report itself fail on a non-CCSDS definition.
+                    if "PKT_APID" in packet:
+                        message += f" APID={packet['PKT_APID']}."
+                    elif isinstance(packet.binary_data, ccsds.CCSDSPacketBytes):
```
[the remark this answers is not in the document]
medley56: Fixed in 53af2bd, in `CCSDSPacketBytes.__str__` rather than with a length guard here. Below `HEADER_LENGTH_BYTES` it now returns `CCSDSPacket Header: (incomplete, 1 of 6 bytes: 07)` instead of indexing the header properties, which also covers the pre-existing trailing-bits warning further down in `parse_bytes`. Covered by `test_ccsds_packet_bytes_str_short_input` and the one-byte case in `test_parse_bytes_unrecognized_packet_ccsds_bytes_without_pkt_apid`.

space_packet_parser/xtce/definitions.py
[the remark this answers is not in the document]
medley56: Added `test_parse_bytes_unrecognized_packet_ccsds_bytes_without_pkt_apid` in 53af2bd. It parses `CCSDSPacketBytes` against the no-`PKT_APID` definition and asserts the message contains the CCSDS header printout and no `APID=` clause, with a full six-byte header and with a single byte. This is the line codecov flagged as uncovered.

CHANGELOG.md
medley56: Rebased onto main at 8c1ae21 and kept both bullets under Unreleased / Fixed, #285 first. The #276 bullet also gained a sentence about the `str(CCSDSPacketBytes)` change in 53af2bd.

space_packet_parser/xtce/definitions.py
medley56: Done in 53af2bd. `CCSDSPacketBytes.__str__` returns a truncated rendering when `len(self) < HEADER_LENGTH_BYTES`; no change at this line, so header-length knowledge stays in `generators/ccsds.py`. Construction is unchanged, so existing callers that build short `CCSDSPacketBytes` are unaffected. Noted in the changelog since the class is public.

medley56: Not extracting a helper. The comment conditions it on the `__str__` fix not being taken, and 53af2bd takes it. What remains duplicated is a one-line `isinstance` check and a one-line f-string append at two sites; a helper would be longer than that and the surrounding message text differs anyway.

PR #284, opened 2026-09-13
Add Zenodo concept DOI and date-released to CITATION.cff

### LINKED ISSUE #280: CITATION.cff has no DOI, so generated citations omit the archived record
## Problem

`CITATION.cff` (at tag `6.2.0`) declares no DOI. There is no top-level `doi:` key and no `identifiers:` block:

```yaml
cff-version: 1.2.0
title: 'space_packet_parser'
type: software
version: '6.2.0'
description: A CCSDS telemetry packet decoding library based on the XTCE packet format description standard.
license: BSD-3-Clause
abstract: ...
authors: ...
maintainers: ...
repository-code: "https://github.com/lasp/space_packet_parser"
url: "https://space-packet-parser.readthedocs.io"
```

The record exists — the README badge links it — but anything that consumes `CITATION.cff` never sees it. GitHub's "Cite this repository" widget, `cffconvert`, and the BibTeX/APA exports it drives all produce a citation whose only locator is the repo URL, with no persistent identifier.

That is the one field a persistent identifier is for. SPD-41a §III.D.vii expects publicly available SMD-funded software to be citable by a persistent identifier; the DOI is minted and correct, it is just not in the citation metadata where a citing tool will look.

## The DOIs, for reference

Confirmed against the Zenodo API (`conceptdoi` is reported identically on both version records):

| DOI | What it is |
| --- | --- |
| `10.5281/zenodo.7735001` | **Concept DOI** — version-agnostic, resolves to the latest release |
| `10.5281/zenodo.22732164` | Version DOI for 6.2.0 |
| `10.5281/zenodo.19392826` | Version DOI for 6.1.2 |

The README badge already links the concept DOI, which is right and needs no change.

## Suggested fix

Add the concept DOI as an identifier. The concept DOI is the better choice here because it needs no maintenance — it keeps resolving to the newest release on its own, whereas a version DOI would have to be bumped by the release process along with `version:`.

```yaml
identifiers:
  - type: doi
    value: 10.5281/zenodo.7735001
    description: Concept DOI resolving to the latest version of space_packet_parser
```

If you would rather have citations pin the exact version, use the top-level `doi:` key with the version DOI instead, and add it to whatever already updates `version:` at release time:

```yaml
doi: 10.5281/zenodo.22732164
```

Doing both is also valid CFF and is what Zenodo's own guidance suggests: `doi:` for the version being cited, `identifiers:` for the concept.

## Adjacent nit, same file

`date-released` is also absent. CFF recommends it, and the same release step that bumps `version:` is the natural place to set it. Worth doing in the same edit if you are in there anyway.

medley56:
# Summary

`CITATION.cff` declared no persistent identifier, so GitHub's "Cite this repository" widget and `cffconvert` produced citations carrying only the repository URL.

This adds the Zenodo **concept DOI** (`10.5281/zenodo.7735001`) via an `identifiers:` block, and adds `date-released` so generated citations carry a year.

Closes #280

## Why the concept DOI and not a version DOI

A version DOI was deliberately rejected. The ordering of the release process makes it unusable:

1. Release checklist **step 3** hand-edits `CITATION.cff`.
2. Release checklist **step 6** pushes the tag.
3. Zenodo mints the version DOI only *after* it archives the tagged release — i.e. after step 6.

So at the moment `CITATION.cff` is edited, the version DOI for the release being prepared **does not exist yet**. A top-level `doi:` key would therefore routinely ship pointing at the *previous* release — a persistent identifier that resolves to the wrong software, which is worse than no DOI at all.

The concept DOI is version-agnostic, always resolves to the newest release, and requires no maintenance. It is the same DOI already linked from the README badge.

## `date-released` and keeping it fresh

Without `date-released`, the BibTeX/APA exports have no `year`, and the year is not derivable from anything else in the file. A citation with a DOI but no year is still defective.

The value is **quoted** (`date-released: "2026-09-13"`) on purpose: the CFF 1.2.0 schema types `date-released` as `{"type": "string", "format": "date"}`. Unquoted, YAML parses `2026-09-13` into a date object, which fails strict validation.

To stop it going stale, release-checklist step 3 in `docs/source/developers.md` — which *already* required a manual edit of `CITATION.cff` — has been widened to cover the new field. This adds no new step to the release process, only a wider existing one.

**Known residual risk, stated plainly:** `scripts/check_metadata.py` does **not** validate `date-released`. It reads only `title`, `description`, `version`, `license`, `authors`, and `maintainers`. So a forgotten `date-released` will go stale silently — nothing in CI will catch it. This was judged acceptable for a date (a minor inaccuracy in a citation year) but would *not* have been acceptable for a DOI, which is a second reason the version DOI was rejected.

## Note for reviewers: pre-existing CFF schema violations (not introduced here)

While verifying this change against the published CFF 1.2.0 schema I found that `CITATION.cff` **does not currently validate**, and did not before this PR either. Validating the unmodified file from `main` produces the identical errors:

- root key `description` is not a valid CFF 1.2.0 key (the CFF field is `abstract`, which the file also has)
- root key `maintainers` is not a valid CFF 1.2.0 key (the CFF analogue is `contact`)
- `orcid` values must be full `https://orcid.org/…` URIs, not bare identifiers
- the required `message` key is absent

Consequence: `cffconvert` refuses to convert the file at all, so the DOI added here will not reach a `cffconvert` BibTeX export until those are fixed. (GitHub's own widget is more lenient.)

I confirmed the additions in this PR are themselves schema-correct by validating a scratch copy with only the pre-existing offenders corrected — that copy validates cleanly and its BibTeX contains both `doi = {10.5281/zenodo.7735001}` and `year = {2026}`.

**I did not fix those pre-existing issues here**, because they are out of scope for #280 and, more importantly, they are coupled to tooling: `scripts/check_metadata.py` reads the `description` and `maintainers` keys directly, so removing or renaming them would break the `check-space-packet-parser-metadata` hook. That deserves its own PR that changes both files together. Happy to open a follow-up issue.

## Verification

- `pre-commit run --all-files` — all hooks pass (prettier does not process `.cff` and left the file untouched)
- `pre-commit run check-space-packet-parser-metadata --all-files` — passes
- `date-released` confirmed to parse as `str`, not `datetime.date`
- Docs build (`make html`) succeeds
- No version bumps; ships under `## [Unreleased]`


medley56:
Two follow-ups on this branch since the PR was opened:

**1. The pre-existing CFF validity problem now has a ticket: #285.**

To be explicit about what is and isn't in this PR: `CITATION.cff` does not validate against the CFF
1.2.0 schema, and did not before this change either — the unmodified file from `main` fails
identically with `Additional properties are not allowed ('description', 'maintainers' were
unexpected)`. That is out of scope here because fixing it requires changing `scripts/check_metadata.py`
in the same commit (it reads both of those keys, and the `check-space-packet-parser-metadata` hook
runs with `always_run: true`). #285 has the details.

**2. Softened the CHANGELOG entry (`bdb783c`).**

The original entry said the DOI would now appear in `cffconvert`'s BibTeX/APA exports. That is not
true today — `cffconvert` rejects the document outright for the reasons above, so it emits nothing at
all. The entry now claims only what actually happens: the metadata carries the identifier and the
release year, and GitHub's "Cite this repository" widget reads both. The `cffconvert` half becomes
true once #285 lands.

The DOI itself is verified correct: validating a corrected scratch copy of the file produces
`doi = {10.5281/zenodo.7735001}` and `year = {2026}` in the BibTeX output, so the additions in this PR
are schema-correct — they are just sitting in a document that has other problems.

PR #286, opened 2026-09-13
Gate CI on benchmark regression thresholds

### LINKED ISSUE #147: Add Documentation of XTCE Parsing Benchmarking
One of the benchmarks we run is the parsing of a complex CTIM XTCE packet definition file. At one point this parsing took 12s and is now down to under 100ms and we want it to stay that way.

Add a discussion of this benchmarking to the Documentation Benchmarking section.

### LINKED ISSUE #287: Enforce the CTIM XTCE definition load time budget in CI
## Context

#147 asked for the XTCE parsing benchmarking to be documented, and #286 does that: `docs/source/benchmarking.md` now explains that `test_benchmark_ctim_xtce_parsing` exists as a regression guard, records the history (loading `ctim_xtce_v1.xml` once took on the order of twelve seconds because each of its 38 concrete containers re-parsed their shared abstract base from scratch), and states the budget — loading that definition should take tens of milliseconds, and approaching 100 ms means something has regressed. It currently measures around 34 ms in the devcontainer.

What #286 deliberately did **not** do is make any of that true automatically.

CI runs `pytest --color=yes --cov --cov-report=xml`, which collects and executes `tests/benchmark/` on every matrix combination — and then throws the timings away. Nothing is saved, nothing is compared, nothing is gated. `.benchmarks/` exists but is empty and gitignored. A performance regression of any size, including another two-orders-of-magnitude one, will produce a fully green build.

This is a genuine trap rather than a cosmetic gap: a reader who sees a benchmark suite running in CI will reasonably assume it is a gate. #286 documents the gap honestly in a `{note}` and tells contributors touching `space_packet_parser/xtce/` to run `--benchmark-autosave` / `--benchmark-compare` manually, but a manual step that depends on remembering to take it is not a guard. This issue is for designing a real one.

## Driving Requirements

- A regression in XTCE definition load time on the scale of the one #147 refers to should fail a build rather than merge silently.
- Enforcement must not produce intermittent failures on unchanged code. A docs/CI change that makes the build flaky is worse than the status quo, because it trains people to re-run red builds.
- Whatever threshold is chosen should be maintainable — it should not need hand-editing every time a runner image or Python version changes.

## Implementation Requirements

Two candidate approaches, both with real problems that need resolving before anything is implemented.

**Option A — a hard assert in the benchmark test.**

Add something like `assert benchmark.stats.stats.mean < 0.1` to `test_benchmark_ctim_xtce_parsing`.

- Cheap, obvious, no new infrastructure, and the threshold lives next to the thing it guards.
- But it is evaluated on every cell of the CI matrix: Windows, macOS, and five Python versions, on shared GitHub-hosted runners with noisy neighbours and no performance SLA. The 100 ms line has roughly 3x headroom over the devcontainer measurement, which sounds generous until a cold, contended Windows runner is involved. Needs actual data — collect observed means per matrix cell across a number of runs before picking any number, rather than assuming 100 ms transfers.
- A single slow round can drag the mean; consider asserting on the median or minimum instead, since the minimum is the most noise-resistant statistic pytest-benchmark reports.
- May need to be restricted to one matrix cell (a single Linux + pinned Python job) rather than all of them.

**Option B — `--benchmark-compare-fail` against a stored baseline.**

Run `pytest tests/benchmark/ --benchmark-compare --benchmark-compare-fail=mean:25%` in CI.

- Catches relative regressions, which is the thing actually worth catching, and adapts to hardware instead of hardcoding an absolute number.
- But it needs a baseline to compare against, and that is the unsolved part. Baselines are only meaningful when compared against runs from the same machine — ephemeral CI runners are, by design, not the same machine. Options worth evaluating: committing baselines to the repo (churn, merge conflicts, and they still encode one machine's characteristics), caching them via `actions/cache` (evictable, and the first run after an eviction has nothing to compare against), or generating the baseline in the same workflow run by also benchmarking the merge base (doubles benchmark runtime but makes the comparison genuinely same-machine — probably the most promising variant).
- The percentage threshold needs the same empirical treatment as Option A's absolute one. Run-to-run variance on the current benchmark is already a few percent in a quiet devcontainer.

A reasonable outcome may be a narrow version of Option A — an absolute assert with a generous threshold, on one matrix cell only — on the grounds that the failure mode actually being guarded against is two orders of magnitude, not twenty percent, so a loose alarm line catches it while staying far away from the noise floor. But that should be a decision made from measurements, not assumed.

## Considerations

- Scope should probably cover the whole `tests/benchmark/` suite, not just CTIM. The packet-parsing benchmarks have the same gap. CTIM is the one with a documented budget and a documented incident behind it, so it is the natural first case.
- If a threshold is added, `docs/source/benchmarking.md` and the `### Benchmarks` section of `docs/source/developers.md` both currently state plainly that no timing threshold is enforced anywhere. Both must be updated, or they become actively misleading.
- Benchmarks currently run inside the coverage-instrumented pytest invocation. Coverage tracing inflates and distorts timings, so any enforcement should probably run them in a separate, uninstrumented step (`--benchmark-only`, no `--cov`), with the main test job skipping them via `--benchmark-skip`. Worth confirming how much coverage is currently perturbing the numbers.
- Consider whether a failure should be blocking or advisory to begin with. Running the comparison and reporting it without failing, for a few weeks, would produce the per-runner variance data that both options need in order to pick a defensible threshold.

Related: #147, #286

medley56:
# Summary

CI now fails the build when a benchmark in `tests/benchmark/` regresses. Every benchmark test asserts its mean run time against a threshold declared next to it, enforced on `ubuntu-latest` for each supported Python version (3.10 to 3.14).

## How the gate works

- **Thresholds are in seconds on a nominal machine, not raw seconds.** GitHub hands the same job to runners whose speed differs by up to ~2x from run to run, and that difference is common to every benchmark in the job. A `machine_speed_factor` fixture in `tests/benchmark/conftest.py` times a fixed pure-Python reference workload immediately before each benchmark and the threshold is scaled by it. This removed most of the run-to-run swing in the means (Linux CV fell from 10 to 35% to 1 to 12%).
- **Linux only.** macOS runners drift in speed within a job and stall for 100 to 200 ms mid-round, and Windows shows the same bimodal hardware classes as Linux with more residual noise. Both would need margins of 2x or more to avoid false positives, which defeats the purpose, so the gate runs only on `ubuntu-latest`. The other OSes still run the benchmark tests functionally under `--benchmark-disable`.
- **Explicit opt-in.** The check is enforced only when `SPP_BENCHMARK_GATE` is set, which the CI gate step does. A plain `pytest tests/benchmark/` locally just reports timings, so developers on other hardware are not gated against x86-runner thresholds.
- **Calibration.** Thresholds come from the normalized means of six CI runs per Python version, set at about 1.5x the average and never below 1.2x the largest value observed. A single `"default"` per test covers all versions; a version-specific key can be added if one ever needs it. At 1.25x margins two of three validation runs produced one false positive each (a runner-class shift and a burst of slow rounds), which is why the margin is 1.5x.

## CI changes

- The coverage-instrumented test step passes `--benchmark-disable`, since coverage tracing distorts timings.
- A separate uninstrumented "Benchmark regression gate" step runs `tests/benchmark/` on `ubuntu-latest` with `SPP_BENCHMARK_GATE=1`.
- Each Linux job writes the per-test normalized mean and speed factor to the step summary and uploads `benchmark.json` as a `benchmark-py<version>-attempt<n>` artifact, so recalibration does not require log scraping.
- `codecov.yml` ignores `tests/benchmark/`, which by design never runs instrumented.

## Benchmark changes

- The five bit-reading benchmarks run 1000 `pedantic` rounds instead of 3, so a stall of a few milliseconds moves the mean by a few percent rather than twenty.
- The two packet-parsing benchmarks let pytest-benchmark calibrate the round count instead of a fixed 20.

## Documentation

- `developers.md` gains a short `### Benchmarks` section describing the gate and how to recalibrate.
- `benchmarking.md` gains a `### Regression Guard` note explaining the fixed bug the CTIM benchmark guards against and pointing at the developer docs.
- CHANGELOG entry under `[Unreleased]`.

## Verification

- All 9 benchmarks pass locally with the gate off, on, and under `--benchmark-disable`; forcing an absurd threshold fails the test.
- CI run #549 (same thresholds) passed on all five Linux jobs; the squashed commit re-runs the full matrix.
- `make html` builds with no new warnings.

Closes #147
Closes #287


docs/source/developers.md
```diff
@@ -13,6 +13,25 @@ To run all tests, run
 pytest tests
 ```
 
+### Benchmarks
+
+`tests/benchmark/` holds a [pytest-benchmark](https://pytest-benchmark.readthedocs.io/) suite
+covering XTCE definition loading, full packet parsing, and the low level bit-reading primitives.
+These run as part of the normal test suite and in CI, but no timing threshold is enforced anywhere —
+a performance regression will not fail a build.
```
medley56: This is really what we should do. Documentation is not harmful, but the real solution to this is to check benchmarks against thresholds. If a set threshold is exceeded, that should fail the build. This changes the scope of this PR but that is OK. We should implement the real solution.

docs/source/benchmarking.md
```diff
@@ -215,6 +215,84 @@ The practical advice holds either way: load a definition once and reuse the resu
 {py:class}`~space_packet_parser.xtce.definitions.XtcePacketDefinition` across all the files you
 parse, rather than re-reading the XTCE document per file or per process.
 
+### The CTIM Definition as a Regression Guard
```
medley56: This is too verbose. We should rely on CI tests checking this regression state, not a document that likely no one will read. Again, this is changing the scope of the PR but it's clear that we should actually check this in CI.

medley56 (commented):
Ready to merge

PR #290, opened 2026-09-13
Fix CFF citation schema and metadata consistency checks

### LINKED ISSUE #285: CITATION.cff does not validate against the CFF 1.2.0 schema
# Context

Found while adding the Zenodo DOI in #280 / [#284](https://github.com/lasp/space_packet_parser/pull/284).
`CITATION.cff` has never validated against the CFF 1.2.0 schema it declares on line 1. This is
independent of that PR — the unmodified file from `main` fails identically:

```console
$ git show main:CITATION.cff > /tmp/citation_main.cff
$ uvx cffconvert --validate -i /tmp/citation_main.cff
jsonschema.exceptions.ValidationError: Additional properties are not allowed
('description', 'maintainers' were unexpected)
```

Because the document is rejected outright, `cffconvert` cannot produce BibTeX or APA output at all.
GitHub's "Cite this repository" widget is more lenient and still renders a citation, which is
presumably why this has gone unnoticed.

## The violations

| Problem | Detail |
| --- | --- |
| `description` is not a CFF key | The schema is `additionalProperties: false` at root. The CFF field for this is `abstract`, which the file **already has** — so `description` is redundant as well as invalid. |
| `maintainers` is not a CFF key | The CFF analogue is `contact`. |
| `orcid` format | Must be a full `https://orcid.org/0000-0002-3520-9715` URI, not the bare `0000-0002-3520-9715`. |
| `message` is missing | It is a **required** key. Conventionally: `message: "If you use this software, please cite it as below."` |

# Driving Requirements

`CITATION.cff` should validate against CFF 1.2.0, so that `cffconvert` and the BibTeX/APA exports it
drives actually work, and so the DOI added in #280 reaches the tools that consume this file.

# Implementation Requirements

- Add the required `message` key.
- Replace the root-level `description` with the existing `abstract` (verify the abstract text is what
  you want the citation to carry, then delete `description`).
- Rename `maintainers` to `contact`.
- Expand both `orcid` values to full `https://orcid.org/...` URIs.
- Verify with `uvx cffconvert --validate -i CITATION.cff`, and confirm
  `uvx cffconvert -f bibtex -i CITATION.cff` emits an entry containing the DOI.

# Considerations

- **This is coupled to tooling and cannot be a pure CFF edit.** `scripts/check_metadata.py` reads
  `citation_cff["description"]` and `citation_cff["maintainers"]` directly to cross-validate against
  `pyproject.toml` and `meta.yaml`. Removing or renaming those keys **will break the
  `check-space-packet-parser-metadata` pre-commit hook**, which runs with `always_run: true`. Both
  files must change in the same PR.
  - For `description`: point the check at `abstract`, or at whichever field should be authoritative.
    Worth deciding deliberately — `pyproject.toml`'s `description` is a one-liner while `abstract` is
    a paragraph, so they may not be comparable and the check may need to drop that comparison.
  - For `maintainers`: the check compares `{(name, email)}` sets; `contact` has the same shape, so
    this should be a rename.
- Adding a `cff-validator` pre-commit hook would stop this recurring, but it needs network access on
  every commit. Probably a separate discussion — note that `validate-codecov-yaml` already sets the
  precedent of a network-dependent hook listed in `ci.skip`.
- No urgency: the citation still renders on GitHub today. This is about the file being correct for
  every other consumer.

Gambit-Checkmate:
# Summary

`CITATION.cff` is rejected by the CFF 1.2.0 schema, preventing citation exports. Add the required citation message, replace `maintainers` with `contact`, expand ORCID identifiers to full URLs, and remove the unsupported `description` field while preserving the existing abstract.

Update the metadata consistency check to read `contact`. Package descriptions continue to be compared between `pyproject.toml` and `meta.yaml`; the longer citation abstract is intentionally independent.

Closes #285.

## Validation

- Confirmed that `cffconvert --validate -i CITATION.cff` failed on the original file and passes after the change.
- BibTeX and APA-like exports both succeed.
- Verified separately that adding the DOI identifier proposed in #284 to the corrected metadata produces a valid BibTeX entry containing `10.5281/zenodo.7735001`. This PR does not duplicate the DOI/date changes from #284.
- `python scripts/check_metadata.py` passes. Deliberately mismatching a package description or contact email still raises the expected consistency error.
- `pytest --benchmark-disable -q`: 481 passed (Windows, Python 3.12).
- Ruff lint/format and `git diff --check` pass.
- All applicable pre-commit hooks pass on the three changed files, including the always-run metadata check.

AI assistance was used to prepare this contribution and run the checks above.


CITATION.cff
```diff
@@ -18,16 +18,16 @@ abstract: The Space Packet Parser Python library is a generalized, configurable
 authors:
 - email: gavin.medley@lasp.colorado.edu
   name: Gavin Medley
```
medley56, line 20: 💡 `SUGGESTION` — authors validate as CFF entities, not persons

Every entry under `authors:` and `contact:` uses a single `name:` key. In CFF 1.2.0, `name` belongs to the *entity* (organisation) definition; a *person* is written as `given-names`/`family-names` and has no `name` key. The file passes validation only because the schema accepts `anyOf [person, entity]` for each author, so consumers that distinguish the two (Zenodo's CFF import, APA output that needs a surname to invert) see three corporate authors.

Pre-existing and out of scope for #285: `verify_authors` and `verify_maintainers` in `scripts/check_metadata.py:85,96` both key on `a["name"]`, so switching is not a one-line edit. Flagging it so it can be tracked as a follow-up issue rather than fixed here.

medley56 (approved):
Does exactly what #285 asks and nothing more. I re-verified the schema claims against the CFF 1.2.0 `schema.json` (`message` is required; `contact` is a valid root key; `description` and `maintainers` are not; `orcid` must match `https://orcid.org/…`), confirmed `cffconvert --validate` fails on `main` and passes on this branch, and ran `python scripts/check_metadata.py` clean on the head checkout. The one substantive observation is pre-existing and left as a suggestion for a follow-up issue.

Coordination note: #284 also adds under `## [Unreleased]` in `CHANGELOG.md`, so whichever of the two merges second will need a trivial rebase.

0 critical · 0 warning · 1 suggestion · 0 nitpick

PR #292, opened 2026-09-13
Add frontmatter to the update-changelog skill and trim it down

### LINKED ISSUE #281: Add frontmatter to the update-changelog skill and trim it down
# Context

`.claude/skills/update-changelog/SKILL.md` predates the skill format we now use. Two problems
surfaced while cutting the 6.2.0 release:

1. **It has no YAML frontmatter.** Claude Code discovers skills by reading the `name` and
   `description` keys in a frontmatter block at the top of `SKILL.md`. This file starts straight in
   at `Update CHANGELOG.md based on recent git history.`, so the skill is very likely invisible to
   the agent — it only gets used if a human points at the file directly. The `cut-a-release` skill
   added in #278 has the frontmatter block for comparison.

2. **It is considerably longer than it needs to be.** At ~100 lines it spells out things the agent
   already knows or can see for itself: the full Keep a Changelog category list with glosses, the
   exact markdown of a two-line heading insert, a Step 5 checklist that restates constraints already
   stated in Step 4. Skill files are loaded into context every time the skill fires, so length is a
   real cost, and padding dilutes the instructions that actually matter.

# Driving Requirements

The skill should be discoverable, and should contain the minimum needed to convey the process to an
agent — nothing more.

# Implementation Requirements

- Add a frontmatter block with `name: update-changelog` and a `description` that says both what the
  skill does and when to use it (the description is the only thing the agent sees when deciding
  whether to invoke it, so it needs the trigger conditions, not just the summary).
- Cut the file down. Keep what is genuinely repo-specific and non-obvious:
  - the three files that carry the version (`pyproject.toml`, `meta.yaml`, `CITATION.cff`) and that
    they must agree
  - the Scenario A / Scenario B branch and the question to ask to tell them apart
  - the footer diff-link bookkeeping, which is fiddly and easy to get wrong
  - the entry format with the issue link
  - skipping Dependabot bumps and CI-only changes
- Drop what an agent already knows or can read off the file it is editing: the Keep a Changelog
  category glosses, the illustrative markdown snippets for trivial edits, and the Step 5
  verification checklist.
- Confirm afterwards that the skill still actually fires when asked to update the changelog.

# Considerations

- **There is a latent bug in Step 1 worth fixing while the file is open.** It instructs
  `git log <LATEST_TAG>..HEAD` to find unreleased work. That is unreliable in this repo: release tags
  live on long-lived `release/X.Y` branches that carry only cherry-picked fixes, so the range lists
  commits that already shipped. This is not hypothetical — it is how XTCE `UnitSet` support (#47,
  PRs #258/#261) reached `main` with no changelog entry and was nearly missed in 6.2.0. The reliable
  test for whether a commit is already released is
  `git merge-base --is-ancestor <sha> <tag>`. The `cut-a-release` skill documents this trap; this one
  should too, and that note is worth the lines it costs.
- Trimming should not become rewriting. The process the file describes is correct and was followed
  successfully for 6.2.0 — this is an edit for length and discoverability, not a redesign.
- Whoever picks this up should skim the other skills in `.claude/skills/` at the same time; if any
  others are missing frontmatter, they have the same discoverability problem.

burakeyler:
Closes #281

## What

- **Frontmatter.** `.claude/skills/update-changelog/SKILL.md` now starts with `name: update-changelog` and a `description` that states both what it does and when to use it (update / fill in / catch up the changelog, or while preparing a release), so Claude Code can discover it.
- **Trimmed ~110 → 64 lines.** Kept the repo-specific parts listed in the issue: the three version files and that they must agree, the Scenario A/B branch and the question that decides it, the footer diff-link bookkeeping, the entry format with the issue link, and skipping Dependabot/CI-only changes. Dropped the Keep a Changelog glosses, the snippets for trivial edits, and the Step 5 checklist.
- **Step 1 trap fixed.** Documents that `git log <LATEST_TAG>..HEAD` includes already-shipped commits because tags live on `release/X.Y`, and uses `git merge-base --is-ancestor <sha> <tag>` as the check, matching the wording in the `cut-a-release` skill from #278. Also notes that `[Unreleased]` may be incomplete.

The process itself is unchanged — this is a length/discoverability edit, not a redesign.

## Other skills

`update-changelog` is the only skill on `main` today; `cut-a-release` (#278) already has frontmatter.

## Verification

- Parsed the frontmatter with PyYAML: valid, keys `name` and `description`.
- I haven't been able to confirm in your environment that the skill fires on "update the changelog"; a quick check by a maintainer with Claude Code after merge would close the last acceptance item.


.claude/skills/update-changelog/SKILL.md
```diff
@@ -1,99 +1,64 @@
-Update `CHANGELOG.md` based on recent git history. Follow these steps exactly.
+---
+name: update-changelog
+description: Update CHANGELOG.md from the git history since the last release — add Keep a Changelog entries for unreleased PRs, or turn [Unreleased] into a versioned section when the version has been bumped. Use when asked to update, fill in, or catch up the changelog, or while preparing a release.
```
medley56: 💡 `SUGGESTION` — drop "or while preparing a release" from the description

Please narrow this to changelog-only triggers. Once #278 merges, "prepare a release" matches this description and `cut-a-release`'s "Use when asked to cut, prepare, or publish a release" equally well. The two are not composable — `cut-a-release` does the changelog work inline in its Step 5 rather than delegating here — so an agent that resolves "prepare a release" to this skill updates the changelog and skips the version bump and the tag. I would rather `cut-a-release` own the release phrasings outright. Nothing else about the frontmatter needs changing.

```suggestion
description: Update CHANGELOG.md from the git history since the last release — add Keep a Changelog entries for unreleased PRs, or turn [Unreleased] into a versioned section when the version has been bumped. Use when asked to update, fill in, or catch up the changelog.
```

.claude/skills/update-changelog/SKILL.md
```diff
@@ -1,99 +1,64 @@
...
+---
 
-## Step 1 — Gather context
+# Update the changelog
 
-Run the following commands:
+## Step 1 — Gather context
```
medley56: 💡 `SUGGESTION` — say which branch this runs on

Please add the guard line below. Every command in this step resolves against the local `HEAD`, and the edit in Step 4 lands in the local `CHANGELOG.md`. The new `git fetch origin --tags` refreshes the tag list but not the checkout, so on a stale `main` or a feature branch the result is a changelog covering the wrong set of commits, and no later step would catch it. The old file was equally silent, so this is not something you introduced — but adding the fetch is what makes it a half-measure rather than a plain omission, which is why I would like it closed here.

```suggestion
## Step 1 — Gather context

Run this on an up-to-date `main`: the commands below read the local `HEAD` and Step 4 edits the local
`CHANGELOG.md`. If the checkout is on another branch or behind `origin/main`, say so and stop.

```

.claude/skills/update-changelog/SKILL.md
```diff
@@ -1,99 +1,64 @@
...
+tip of `main` in the usual way. Before writing an entry for a commit, confirm it is unreleased:
 
 ```bash
-git log <LATEST_TAG>..HEAD --oneline --no-merges
-git log <LATEST_TAG>..HEAD --oneline --merges
+git merge-base --is-ancestor <sha> "$LATEST_TAG" && echo "already released" || echo "unreleased"
 ```
 
-Read these files in full:
+Do not assume `[Unreleased]` is complete either — features have reached `main` without an entry.
```
medley56: ⚠️ `WARNING` — the Step 1 check can never report "already released"

Please replace this whole block with the suggestion below. Four separate things are wrong with the mechanism as written, and they are cheaper to fix in one edit than four:

1. **The check is a tautology.** `git log "$LATEST_TAG"..HEAD` and `git merge-base --is-ancestor <sha> "$LATEST_TAG"` are complements by definition: `A..HEAD` means "reachable from HEAD, not reachable from `A`", and "ancestor of `A`" means "reachable from `A`". Every SHA the first command prints is guaranteed to fail the second. On this repo, 0 of the 60 commits in `6.0.1..origin/main` and 0 of the 3 in `6.2.0..origin/main` come back flagged.
2. **`$LATEST_TAG` is empty in the second block.** Each Bash call is a fresh shell, so the variable assigned in the first block does not survive. `git merge-base --is-ancestor <sha> ""` errors, `||` swallows it, and the output is `unreleased` again — it fails open in the same direction.
3. **The topology claim is false for the tags this will actually meet.** Release branches get merged back, so `6.2.0`, `6.1.2`, `6.1.1` and `6.1.0` are all ancestors of `main` — `6.2.0` sits at `8a43eb1`, reachable via `e6359c6` ("Merge pull request #277 from lasp/release/6.2"). Only `6.0.1` and older match the shape described here. The condition is real but intermittent, so it should be stated as a thing that can happen, not a property of the repo.
4. **`--sort=-version:refname` ranks prereleases above their release.** `6.0.0rc4` outranks `6.0.0` in our tag list, so an unprefixed rc tag would silently become `LATEST_TAG`. Pre-existing, but this line is being rewritten anyway.

The signal you want is cherry-pick equivalence, not ancestry: `--cherry-mark` marks with `=` those commits whose patch already exists on the tagged side, which is exactly the case where a fix was cherry-picked onto `release/X.Y` and shipped under a different SHA. I checked that it keeps merge commits, which matters because Step 3 reads PR numbers out of "Merge pull request #NNN" subjects. Folding everything into one code block also removes problem 2 for free.

````suggestion
```bash
git fetch origin --tags
# Skip prerelease tags: `--sort=-version:refname` ranks 6.0.0rc4 above 6.0.0.
LATEST_TAG=$(git tag --sort=-version:refname | grep -E '^[0-9]+\.[0-9]+(\.[0-9]+)?$' | head -1)
git log --oneline --cherry-mark --left-right "$LATEST_TAG"...HEAD
```

Read the markers, not just the subjects. `>` is on `main` only and is a candidate for an entry; `=`
means the same patch already shipped under `$LATEST_TAG` and must be skipped. A release is tagged on
its `release/X.Y` branch, so a fix cherry-picked onto one ships under a different SHA than its copy
on `main`, and `=` is what catches that. Keep merge commits — Step 3 reads PR numbers out of their
subjects.

Do not assume `[Unreleased]` is complete either — features have reached `main` without an entry.
````

File: .claude/skills/update-changelog/SKILL.md [hunk 4]
```diff
@@ -1,99 +1,64 @@
...
 
-- `CHANGELOG.md`
-- `pyproject.toml` (find `version =` under `[project]`)
-- `meta.yaml` (find `version:` field)
-- `CITATION.cff` (find `version:` field)
+Read `CHANGELOG.md` and the version in all three metadata files: `pyproject.toml` (`[project]`
+`version`), `meta.yaml` (`version:`) and `CITATION.cff` (`version:`). The three must agree; if they
+do not, stop and tell the user.
```
medley56: ✏️ `NITPICK` — `meta.yaml`'s version is nested under `package:`

Please name the nesting, as `cut-a-release` does. It is currently the only `version:` key in `meta.yaml`, so a search finds it either way and nothing is broken today — but the other two paths here are given precisely and this one is not, and a second `version:` appearing in that file later would make the difference matter. Carried over from the old file rather than introduced here.

```suggestion
Read `CHANGELOG.md` and the version in all three metadata files: `pyproject.toml` (`[project]`
`version`), `meta.yaml` (`package:` → `version:`) and `CITATION.cff` (`version:`). The three must
agree; if they do not, stop and tell the user.
```

.claude/skills/update-changelog/SKILL.md
```diff
@@ -1,99 +1,64 @@
...
 
-## Step 2 — Detect release scenario
+## Step 2 — Pick the scenario
 
-Compare the version in `pyproject.toml` / `meta.yaml` / `CITATION.cff` against `LATEST_TAG`.
+- **Version in the files differs from `LATEST_TAG`:** ask the user "Version X.Y.Z is in
+  pyproject.toml/meta.yaml/CITATION.cff but the latest tag is LATEST_TAG. Is this a new release that
+  should get its own versioned section?" Yes → Scenario A. No → Scenario B.
```
medley56: 💡 `SUGGESTION` — name `AskUserQuestion` here rather than "ask the user"

Please put the tool name back. This is the skill's only decision point, and it gates whether the agent rewrites a heading into a versioned release section or only appends to `[Unreleased]`; getting it wrong in the Scenario A direction fabricates a release section for a version that was never tagged. A prose "ask the user" is satisfiable by asking rhetorically and continuing, which is a real failure mode in a non-interactive run. Naming the tool costs three words and makes the stop enforceable. #281 asked for "the question to ask to tell them apart" to be kept, and the tool is part of that.

```suggestion
- **Version in the files differs from `LATEST_TAG`:** use `AskUserQuestion` to ask "Version X.Y.Z is
  in pyproject.toml/meta.yaml/CITATION.cff but the latest tag is LATEST_TAG. Is this a new release
  that should get its own versioned section?" Yes → Scenario A. No → Scenario B.
```

.claude/skills/update-changelog/SKILL.md
```diff
@@ -1,99 +1,64 @@
...
-- `PREV_TAG` = `LATEST_TAG`
 
-1. Rename the `## [Unreleased]` heading to `## [NEW_VERSION] - TODAY`. Populate its body with the categorized entries from Step 3, preserving any entries that were already in `[Unreleased]` before them.
-2. Insert a new empty `## [Unreleased]` section above the renamed section:
+## Step 4 — Edit `CHANGELOG.md`
 
-   ```markdown
-   ## [Unreleased]
+**Scenario A — new release.** Rename `## [Unreleased]` to `## [NEW_VERSION] - YYYY-MM-DD` (today),
+keeping its existing entries ahead of the new ones, and add an empty `## [Unreleased]` above it.
```
medley56: ✏️ `NITPICK` — define `NEW_VERSION` before using it

Please add the definition. The old file had "`NEW_VERSION` = version from `pyproject.toml`"; Scenario A now uses the token four times — once in the heading, three times in the footer snippet — without saying where it comes from. It is inferable from Step 1, which reads all three files and requires them to agree, but it is the one item in the entire trim that was lost rather than dropped deliberately or restated elsewhere, so I would rather close it than rely on the inference.

```suggestion
**Scenario A — new release.** `NEW_VERSION` is the version read from `pyproject.toml` in Step 1.
Rename `## [Unreleased]` to `## [NEW_VERSION] - YYYY-MM-DD` (today), keeping its existing entries
ahead of the new ones, and add an empty `## [Unreleased]` above it.
```

.claude/skills/update-changelog/SKILL.md
```diff
@@ -1,99 +1,64 @@
...
+Then fix the footer links, which are easy to get wrong:
```
medley56: ⚠️ `WARNING` — nothing here says to replace the existing `[unreleased]` line

Please restore the verbs, as in the suggestion. The old text carried one for each line: "Change the `[unreleased]` link to" and "Insert a new versioned link immediately after the `[unreleased]` line". Read literally, this block is now a thing to add. An agent that appends both lines leaves two `[unreleased]:` definitions in the footer; CommonMark takes the first, so the stale one pointing at the previous tag silently wins, and nothing renders as a duplicate to give the mistake away. This is the bookkeeping #281 specifically asked to keep, so it is worth the extra clause.

```suggestion
Then fix the footer links, which are easy to get wrong. Replace the existing `[unreleased]` line
with the first of these, and insert the second directly below it:
```

medley56 (changes requested):
Thanks for this — the frontmatter is right (valid YAML, `name` matches the directory, and the trigger phrasings cover the realistic asks), and I checked the central claim rather than taking it on trust. I walked all 34 instructions, paths, commands and edge cases in the old file against the new one: the trim is faithful. Every item is preserved, reworded without changing behavior, or dropped deliberately, including all five Step 5 checklist constraints, which are each restated elsewhere in the new file. "The process itself is unchanged" holds for the trim, which is most of the diff.

It does not cover Step 1, which is new text rather than shortened text, and that is where the two warnings are. The `merge-base --is-ancestor` check cannot ever report "already released" for a commit taken from the range it is meant to filter — the two commands are complements by construction — and the topology it rests on is false for the four most recent tags. Neither is your invention; the wording comes from #281 and from `cut-a-release` in #278, and I will fix #278 on my side. But this PR is what lands it on `main`, so I would like it corrected here.

I'll handle the updates on my side.

2 warnings, 3 suggestions, 2 nitpicks. No critical findings.

medley56:
@burakeyler — I've taken this over rather than sending seven suggestions back to you. All of them are applied in `997e82c`, and the branch is rebased onto `fa65e8c`.

To be clear about what was wrong: the Step 1 mechanism came from #281 and from `cut-a-release` in #278, not from you. I'm correcting #278 separately. The frontmatter — which is what #281 actually asked for — was right as submitted and is unchanged apart from dropping the release phrasing.

Your open acceptance item is closed. With this file in place, Claude Code lists the skill as `update-changelog` with your description, so discovery works.

.claude/skills/update-changelog/SKILL.md
medley56: Applied in `997e82c`. The description now ends at "catch up the changelog", leaving the release phrasings to `cut-a-release`.

.claude/skills/update-changelog/SKILL.md
medley56: Applied in `997e82c`, with the paragraph wrapped one word earlier than in my suggestion so the file stays inside its 100-column wrap. The text is unchanged.

.claude/skills/update-changelog/SKILL.md
medley56: Applied in `997e82c`, with one clause added beyond what I suggested: the paragraph now also explains `<`. `--left-right` emits three markers and my block covered two. It does not arise against the tag the new filter picks today — `6.2.0...HEAD` is three commits, all `>` — but against `6.0.1` there are four (`0da3e6d`, `8540a65`, `652fa6e`, `f32bc68`), all release-branch-only commits that must not get an entry. An unexplained marker in output the file tells you to read was a live gap on an older tag.

I ran the block as it now stands in a single shell: it selects `6.2.0`, prints the three `>` commits, and keeps the merge commit `e6359c6` that Step 3 reads the PR number out of.

One thing worth recording for whoever reads this later — I could not make `=` fire anywhere in this history. There are no cherry-picked duplicates between `main` and any release branch today, so the mechanism is correct for the case it describes but has not been exercised against a real occurrence here.

.claude/skills/update-changelog/SKILL.md
medley56: Applied in `997e82c` — now `package:` → `version:`, matching how `cut-a-release` names the path.

.claude/skills/update-changelog/SKILL.md
medley56: Applied in `997e82c`. Step 2 now says "use `AskUserQuestion` to ask", so the stop is enforceable rather than rhetorical.

.claude/skills/update-changelog/SKILL.md
medley56: Applied in `997e82c`. Scenario A now opens by defining `NEW_VERSION` as the version read from `pyproject.toml` in Step 1.

.claude/skills/update-changelog/SKILL.md
medley56: Applied in `997e82c`. I left the "goes directly under `[unreleased]`, keeping versioned links in descending order" sentence below the snippet in place — the new clause restates the placement, but that sentence is still the only statement of the ordering rule.

